PR #30, opened 2020-10-22
Add br and hr element helpers

oderwat:
I was playing around with gomponents as I like this approach and found that HR and BR tags are not yet supported. As they need to have attributes and there is no difference between attributes and children elements I added them but made them into an extra file. I wonder if attributes and child elements should not be handled as separate parameters.

My tests are testing a bit more than just the plain elements. Not sure if you like that. Having a bigger scope for the tests makes me more confident in having a good test coverage of the actual usage. But then it may fail for code outside of the tested scope. What do you think about that?

Currently, there is also no way to enforce HTML rules like having LI inside UL or OL and other stuff. I think it would be cool if this would be enforced by the type checker or as a "strict" mode at runtime.

But I find it a very nice idea as is already. I dislike adding raw HTML code inside the source and I also dislike mixing HTML (or template) code with dynamic code like if/then/else or loops.


markuswustenberg:
Hey @oderwat, thanks for your contribution! 😊

> I was playing around with gomponents as I like this approach and found that HR and BR tags are not yet supported.

Yeah, I still have a backlog of supported components. I want to implement most of the components at https://developer.mozilla.org/en-US/docs/Web/HTML/Element , at least the ones that are the most used. Remember that you can always just use `g.El` for components that don't exist as helpers.

> As they need to have attributes and there is no difference between attributes and children elements I added them but made them into an extra file. I wonder if attributes and child elements should not be handled as separate parameters.

Yeah, I thought about that as well. The problem is that you can't have multiple variadic parameters in a function, like: `func someElement(attrs ...g.Attr, children ...g.Node)`, and I quite like the simple variadic parameter design.

> My tests are testing a bit more than just the plain elements. Not sure if you like that. Having a bigger scope for the tests makes me more confident in having a good test coverage of the actual usage. But then it may fail for code outside of the tested scope. What do you think about that?

I think that's fine.

> Currently, there is also no way to enforce HTML rules like having LI inside UL or OL and other stuff. I think it would be cool if this would be enforced by the type checker or as a "strict" mode at runtime.

I intentionally have not gone down this path. I have a feeling it's a can of worms, and it would really complicate the code base. If you want this, you could perhaps use an existing checker on the output of `Render`?

> But I find it a very nice idea as is already. I dislike adding raw HTML code inside the source and I also dislike mixing HTML (or template) code with dynamic code like if/then/else or loops.

Thanks. 😊 Same here.

el/solid.go
```diff
@@ -0,0 +1,13 @@
+package el
```
markuswustenberg, line 1: Could you just add these to `content.go`? That's where they're grouped at https://developer.mozilla.org/en-US/docs/Web/HTML/Element .

el/solid_test.go
```diff
@@ -0,0 +1,21 @@
+package el_test
+
+import (
+	g "github.com/maragudk/gomponents"
+	"github.com/maragudk/gomponents/assert"
+	"github.com/maragudk/gomponents/attr"
+	"github.com/maragudk/gomponents/el"
+	"testing"
```
markuswustenberg, line 8: The project's import order is stdlib before internal imports.

markuswustenberg (changes requested):
Looks good, thanks!

el/solid.go
oderwat: Sure!

el/solid_test.go
oderwat: OK. Will respect this in the future.

oderwat:
Stupid question but are you or me expected to resolve the conversation? We let this open to the reviewer (which is mostly me).

el/solid.go
oderwat: I moved them and also got rid of the dependency on `attr`. I think when you test `el` it should still pass if something is wrong with the `attr` package (yeah I am nitpicking here).

markuswustenberg:
> Stupid question but are you or me expected to resolve the conversation? We let this open to the reviewer (which is mostly me).

Not at all stupid. 😊 I would usually say the one doing the fixing, so in this case, you. But it doesn't matter, you can also leave them open. I'm just trying out Github's code review system. :D

markuswustenberg (approved):
😎

oderwat:
Wow I hate it when depending branches get out of control. Let me fix that mess.

oderwat:
OK. Done.

el/solid.go
markuswustenberg: Sounds reasonable.

el/content_test.go
```diff
@@ -37,3 +37,15 @@ func TestP(t *testing.T) {
...
+
+func TestBr(t *testing.T) {
+	t.Run("returns a br element in context", func(t *testing.T) {
+		assert.Equal(t, `<p>Test<br />Me</p>`, el.P(g.Text("Test"), el.Br(), g.Text("Me")))
+	})
+}
+
+func TestHr(t *testing.T) {
+	t.Run("returns a hr element with class", func(t *testing.T) {
+		assert.Equal(t, `<hr class="test" />`, el.Hr(g.Attr("class","test")))
```
markuswustenberg, line 49: Oh, this hasn't been run through gofmt. Just add a space after the comma in `"class","test"`… 😅

oderwat: Sorry. I need to check how I make go fmt automatic when saving in GoLand. Should be fixed now.

oderwat: Yeah. I was editing from another project and this was not caught by my GoLand file watcher which runs `go fmt` on save. The details... Fixed now (and should not happen again).

markuswustenberg:
Excellent, thanks again. 🥳

PR #31, opened 2020-10-22
Some work on the examples

oderwat:
* I removed the `fmt` module from the readme example so it will compile as-is.
* I was bothered with the raw `ul` look of the menu so I added minimal CSS to make it look more like a standard menu.


markuswustenberg (approved):
Nice, thanks. 😊

PR #34, opened 2020-10-27
Add element helpers and refactor

markuswustenberg:
This change adds a lot of element helpers, and refactors:
- helpers into simple, text, and other helpers
- most tests into table-driven tests, so they're easier to read

Thanks to @oderwat for pushing me to improve the tests. 😉


oderwat:
Just an idea as adding all individual elements as tests seem trivial and bloated for what they do: Write one huge test which has all of the elements in one "test page" using a lot of linefeeds to make them on separate lines. So one can simply add them all into that test. When the test fails it could show the line number of the first non-matching line, which would lead to the problem quite reliably.

markuswustenberg:
> Just an idea as adding all individual elements as tests seem trivial and bloated for what they do: Write one huge test which has all of the elements in one "test page" using a lot of linefeeds to make them on separate lines. So one can simply add them all into that test. When the test fails it could show the line number of the first non-matching line, which would lead to the problem quite reliably. Probably too late :)

Yeah, you're probably right. Or something like table-driven tests. I'm usually not a fan of those, but here they might make sense.

That, or auto-generated tests. ;)

PR #38, opened 2020-10-29
Remove fmt.Sprintf call in attribute Render

markuswustenberg:
Just concatenating the strings is much faster.

Before:

```
make benchmark
go test -bench=.
goos: darwin
goarch: amd64
pkg: github.com/maragudk/gomponents
BenchmarkAttr/boolean_attributes-8         	 8194791	       139 ns/op
BenchmarkAttr/name-value_attributes-8      	 5143292	       229 ns/op
PASS
ok  	github.com/maragudk/gomponents	2.841s
```

After:

```
make benchmark
go test -bench=.
goos: darwin
goarch: amd64
pkg: github.com/maragudk/gomponents
BenchmarkAttr/boolean_attributes-8         	16755404	        67.0 ns/op
BenchmarkAttr/name-value_attributes-8      	10208625	       116 ns/op
PASS
ok  	github.com/maragudk/gomponents	2.702s
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,7 @@
-.PHONY: cover lint test
+.PHONY: benchmark cover lint test
+
+benchmark:
+	go test -bench=.
 
 cover:
 	go tool cover -html=cover.out
--- a/gomponents.go
+++ b/gomponents.go
@@ -129,9 +129,9 @@ type attr struct {
 
 func (a *attr) Render() string {
 	if a.value == nil {
-		return fmt.Sprintf(" %v", a.name)
+		return " " + a.name
 	}
-	return fmt.Sprintf(` %v="%v"`, a.name, *a.value)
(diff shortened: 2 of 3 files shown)


oderwat:
Interesting. Did you try that with something else than `%v` which probably has overhead? No experience here with how the Go compiler has implemented this. I am currently "fighting" something similar in a PHP project, where `"$name = $data"` seems to be better optimized than `$name." = ".$data`. At least for multiple fields, because the `""` version allocates only the final size and copies the parts into it.

markuswustenberg:
Yeah, I tried %s as well, but that was the same as %v. 

Normally I wouldn't micro-optimise like this, but given that this will be called a lot, it seemed worth it. 

I'm playing with a version of Render that doesn't build everything in-memory, by the way, but uses io.Writer instead. I still want to keep a simple API, so haven't decided yet.

PR #40, opened 2020-11-02
Add Map function

markuswustenberg:
`Map` makes it easier to build lists of elements without having to iterate.

### DIFF
--- a/examples/simple/simple.go
+++ b/examples/simple/simple.go
@@ -61,11 +61,11 @@ func navbar(props navbarProps) g.Node {
 		{"/foo", "Foo"},
 		{"/bar", "Bar"},
 	}
-	var lis []g.Node
-	for _, item := range items {
-		lis = append(lis, el.Li(el.A(item.path,
-			attr.Classes(map[string]bool{"is-active": props.path == item.path}),
-			g.Text(item.text))))
-	}
+	lis := g.Map(len(items), func(i int) g.Node {
+		item := items[i]
+		return el.Li(
+			el.A(item.path, attr.Classes(map[string]bool{"is-active": props.path == item.path}), g.Text(item.text)),
+		)
+	})
 	return el.Ul(attr.Class("nav"), g.Group(lis))
 }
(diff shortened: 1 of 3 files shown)


oderwat:
Just my thoughts: 

It does not look easier and may actually perform worse, doesn't it? This version breaks the iterator and replaces it with an array lookup. So even if the actual function got inlined (I think) it will be worse than the normal iterator loop. if at all I would have expected that `map` gives the element to the function, so the function can be abstract to the slice. Your implementation also needs the function to be a closure over the iterated object.

I think `map` is something which should be a language feature supported by the compiler and is complemented by `reduce` and other functional components. Then there is `string.Map()` but this has more internal logic to it than a simple `map` of a function. One also could easily use one of the empty interface{} based map implementations here.

markuswustenberg:
> Just my thoughts:

Appreciated.

> It does not look easier and may actually perform worse, doesn't it? This version breaks the iterator and replaces it with an array lookup. So even if the actual function got inlined (I think) it will be worse than the normal iterator loop.

I'm not too worried about performance here, I'm going for readability/usability. `Map` won't be used internally anywhere. But you're probably right.

> if at all I would have expected that `map` gives the element to the function, so the function can be abstract to the slice. Your implementation also needs the function to be a closure over the iterated object.

Yeah, but we can't do this without passing `interface{}`, which I don't like. This is basically a workaround until we have generics. 😅

oderwat:
Well. I wrote a little Haskell in the past and so I always think about `map`. Go seems to me not very well prepared to use this approach. So I would stay away from presenting it in a module because people may think it is more optimized than another approach if there is a special `x.Map`. I love the iterators though. Coming from programming since the 80s I still feel bad for every multiplication when I just could increase a pointer otherwise ;-)

For the generics: I didn't watch any of the generics updates, but to my thinking, it should offer a way to iterate over slices of any type and should not need an implementation by the module at all. So when generics are here, the `x.Map` should get removed. This is why I mentioned, that one could already use an empty interface version of a `map` which later could be replaced by a real generics version.

markuswustenberg:
Yeah, when generics come around, this will probably be removed. See the [sort package](https://golang.org/pkg/sort/) in stdlib for similar approaches to this (from a design point of view). Until then, this little function saves me some typing in my views, so I'll keep it for now. But thanks a lot for the feedback!

PR #47, opened 2020-11-17
Add example with dot imports

markuswustenberg:


oderwat:
Just you know:
image

markuswustenberg:
Yes, there will often be collisions if imported. Maybe I should add a disclaimer in the example. But luckily, it's up to the client to do this, nothing changes in the library of course.

I can see how it can make certain pure view files easier to read, though, so I'm playing around with it. What do you think?

oderwat:
I am still experimenting with my code-base. Actually, I like it but I am not yet sure how to utilize it.

oderwat:
I wondered if it would be possible to avoid collisions with the go-jet package (typesafe SQL Builder). There is just "Time," "Table," and "Raw" colliding. Raw could be replaced with something else and Time, well, never used it :). Table less likely thought as it is in SQL as in HTML pretty important. I am not sure if it would be "cool" to make Raw/Table/Time "optionally" to something like "Literal," "TTable," and "TTime." Probably a stupid idea. But I would like to use both packages as dot imports kind of. Having dot imports in some files and not in others is not really to my liking. Bummer.

Issue #74, opened 2021-05-08
Potential XSS: Values passed to Attr(...) are not escaped by default


philippta:
Currently values passed to the `Attr(name, value)` function are not escaped during rendering by default and impose a potential security risk. These values should be treated as unsafe data, like any other data rendered with the `Text(...)` function.

### Example
```go
Div(g.Attr("foo", `bar" baz="`)).Render(os.Stdout)
```
```html
<!-- Current behavior -->
<div foo="bar" baz=""></div>

<!-- Expected behavior -->
<div foo="bar&#34; baz=&#34;"></div>
```

### Why is this important?

In most cases html attribute like `class="navbar"`, `id="cta-button"`, `name="email"` are statically defined by the developers and do not impose much risk.

However when rendering user supplied data in e.g. a _change profile_ page a user could escape the html attributes by choosing a name like: `Little Bobby"> Tables` which then leads to a potential XSS.

#### Insecure Example
```go
func InputField(name, defaultValue string) g.Node {
    return Input(Name(name), Value(defaultValue)) // calling Attr(...)
}

InputField("firstname", `Little Bobby"> Tables`).Render(os.Stdout)
```

```html
<input name="firstname" value="Little Booby"> Tables">
```

## Proposal

1. Escape all values passed to `Attr(name, value)` by default. (This will _and should_ also affect all other attribute helper functions)
2. Introduce a second, insecure function e.g. `AttrRaw(name, rawvalue)`

## cf.
The `html/template` package has a whole range of escaping functions, including one for [html attributes](https://github.com/golang/go/blob/b38b1b2f9ae710ee2c16a103bb21644f1adbc5d3/src/html/template/escape.go#L63).

markuswustenberg:
Hi @philippta . First of all, thank you for your detailed issue description! This is great, and I value it a lot. 😊

I've been thinking about this as well, and also saw your comment on Reddit about something similar. I agree that it's really cool that `html/template` does this automatically. But I can't really decide whether this is something I should include as well or not. 

As you mention yourself, in most cases attributes are defined by developers and not based on data from elsewhere, and I find that it would perhaps make the API more confusing if I were to introduce different helpers for escaping e.g. attribute values. But I definitely see your point that the current implementation is unsafe in certain threat models. Of course, I could make the default implementation safe against XSS, but I think I'd like to think about the implications about that first. I lean towards making safe the default, though.

[...]

PR #77, opened 2021-05-11
Escape Attr values

markuswustenberg:
Because this can be a place of injection if untrusted data is passed, escape all attribute values.

This slows down rendering a bit, but we're still talking tiny increments compared to everything else.

### Before

```
$ make benchmark
go test -bench=.
goos: darwin
goarch: arm64
pkg: github.com/maragudk/gomponents
BenchmarkAttr/boolean_attributes-8         	13095948	        91.92 ns/op
BenchmarkAttr/name-value_attributes-8      	 9124303	       131.5 ns/op
BenchmarkEl/normal_elements-8              	 6384505	       187.4 ns/op
PASS
ok  	github.com/maragudk/gomponents	5.124s
```

### After

```
$ make benchmark
go test -bench=.
goos: darwin
goarch: arm64
pkg: github.com/maragudk/gomponents
BenchmarkAttr/boolean_attributes-8         	11524586	        92.07 ns/op
BenchmarkAttr/name-value_attributes-8      	 7606478	       157.0 ns/op
BenchmarkEl/normal_elements-8              	 6390315	       187.0 ns/op
PASS
ok  	github.com/maragudk/gomponents	4.000s
```

See #74.

### DIFF
--- a/gomponents.go
+++ b/gomponents.go
@@ -66,8 +66,6 @@ func El(name string, children ...Node) Node {
 	return NodeFunc(func(w2 io.Writer) error {
 		w := &statefulWriter{w: w2}
 
-		name = template.HTMLEscapeString(name)
-
 		w.Write([]byte("<" + name))
 
 		for _, c := range children {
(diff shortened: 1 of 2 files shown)


oderwat:
I think this is the right way to go. XSS is one thing, strange bugs are the other. Automagic escaping really helps to not break stuff by accident.

markuswustenberg:
@oderwat thanks for the feedback. 😊 Have you seen the discussion in #74 as well? I think some of this will still be changed, I'm not sure I'm keeping attribute name escaping.

oderwat:
Yes, I saw it. I actually think that the system should probably "panic" for invalid attribute names. Alternative mangle them into something valid (replacing invalid characters with an _ maybe). Valid could be everything except "whitespace" and `/>"'=`.

markuswustenberg:
> Yes, I saw it. I actually think that the system should probably "panic" for invalid attribute names. Alternative mangle them into something valid (replacing invalid characters with an _ maybe). Valid could be everything except "whitespace" and `/>"'=`.

Hmm. I think I'd rather not panic, because I would be very sad as a library user if the library panicked and I didn't expect it. I think I'll hope that the developer knows what s/he is doing instead, and let them misuse the library. I think that's the least surprising thing to do.

If this turns out to be an issue, I can revisit the decision.

PR #99, opened 2022-05-25
Add IfFunc helper

markuswustenberg:
`IfFunc` is like `If`, except it takes a callback function that returns a `Node` instead of a `Node` directly.
Great when the condition is a nil check and the function uses the value that's not nil.

### DIFF
--- a/gomponents.go
+++ b/gomponents.go
@@ -235,3 +235,12 @@ func If(condition bool, n Node) Node {
 	}
 	return nil
 }
+
+// IfFunc is like If, except it takes a callback function that returns a Node instead of a Node directly.
+// Great when the condition is a nil check and the function uses the value that's not nil.
+func IfFunc(condition bool, cb func() Node) Node {
+	if condition {
+		return cb()
+	}
+	return nil
+}
--- a/gomponents_test.go
+++ b/gomponents_test.go
@@ -291,3 +291,37 @@ func ExampleIf() {
 	_ = e.Render(os.Stdout)
 	// Output: <div><span>You lost your hat!</span></div>
 }
+
+func TestIfFunc(t *testing.T) {
+	t.Run("returns the result of the callback function if condition is true", func(t *testing.T) {
+		n := g.El("div", g.IfFunc(true, func() g.Node { return g.El("span") }))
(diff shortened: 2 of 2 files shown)


markuswustenberg:
@oderwat @averageflow if you're still interested in this project and have some spare time, I would love to hear your thoughts on this little change. 😊

oderwat:
It is cool but I am not sure if it is so much better than a simple if/else. That is because you always have to type the function definition in addition to the construct itself. As this is needed you also can simply use it as the inline function.

But with [go-app](https://go-app.dev/declarative-syntax) (which I actually use now) the "app.If()" is much worse than your idea and I thought about having something similar to your version with go-app but then I ended up just inlining the if. See https://github.com/maxence-charriere/go-app/issues/632

Sadly this is not possible with your package because the anonymous function does not satisfy the interface. I just leave this here

```go
package main

import (
	"os"
	"testing"
)

func IfFunc(condition bool, cb func() g.Node) g.Node {
	if condition {
		return cb()
	}
	return nil
}

// N is needed to implement the interface for anonymous functions
func N(cb func() g.Node) g.Node {
	// delegate to the anonymous function
	return cb()
}

func IfTestA(t *testing.T) {
	var message *string
	e := g.El("div",
		IfFunc(message != nil, func() g.Node {
			return g.El("span", g.Text(*message))
		}),
	)
	_ = e.Render(os.Stdout)
}

func IfTestB(t *testing.T) {
	var messagePtr *string
	e := g.El("div", N(func() (r g.Node) {
		if messagePtr != nil {
			r = g.El("span", g.Text(*messagePtr))
		}
		return
	}))
	_ = e.Render(os.Stdout)
}
```

P.S.: Sadly there also is no way to create a "NotNil(value, func() g.Node)" with generics as there is no `a == T(nil)`.

markuswustenberg:
@oderwat thank you for your detailed reply! You're very kind. 😊

Your `N` function gave me a different idea: to create a function type that is also a `Node`. I'm exploring that in #101. That way, there's no choice between `If`/`IfFunc`, but as you also mention, it's still slightly verbose. Basically, it's a tradeoff between a wrapper type or having double if conditionals.

Any additional thoughts?

PR #101, opened 2022-05-25
Add Lazy type

markuswustenberg:
`Lazy` is a function that returns a `Node` that is itself also a `Node`.

Used with `If`, it enables lazy evaluation of the node to return, depending on the condition passed to `If`.

See also #99 for a different approach explored.


oderwat:
This is a neat trick. I like this kind of lazy evaluation as it also works for if/else and even for something as `MaybeString(v *string,func(v string) g.Node)`

markuswustenberg:
Yeah, I think I like this solution too. Useful anywhere where lazy node evaluation could be nice. Maybe I should just call it `Lazy`.

markuswustenberg:
Hmm. Sadly this doesn't work for attributes. I'll have to rethink a bit.

PR #104, opened 2022-06-03
[feat] Replace `Map` with a generic implementation

codemicro:
Hello! :)

Based off this passage in your introductory blog post:

> Unfortunately, we don't have the handy map function like in Javascript (until we get generics 🤞)
> \- https://www.maragu.dk/blog/gomponents-declarative-view-components-in-go/

I replaced the implementation of `Map` with a generic one that's (in my opinion) easier to use than the previous. The examples and tests have also been updated to match. The CI test runner has also had versions of Go prior to 1.18 removed from the testing matrix.

A couple of things to consider:
 * This is very much a breaking change
 * This bumps the minimum required Go version from 1.15 to 1.18

Thanks!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,8 +18,6 @@ jobs:
     strategy:
       matrix:
         go:
-          - 1.16
-          - 1.17
           - 1.18
 
     steps:
--- a/examples/simple/simple.go
+++ b/examples/simple/simple.go
@@ -58,8 +58,8 @@ func Navbar(currentPath string, links []PageLink) g.Node {
 		Ul(
 			NavbarLink("/", "Home", currentPath),
 
-			g.Group(g.Map(len(links), func(i int) g.Node {
-				return NavbarLink(links[i].Path, links[i].Name, currentPath)
+			g.Group(g.Map(links, func(pl PageLink, _ int) g.Node {
+				return NavbarLink(pl.Path, pl.Name, currentPath)
 			})),
 		),
 
--- a/examples/tailwindcss/tailwindcss.go
+++ b/examples/tailwindcss/tailwindcss.go
@@ -81,8 +81,8 @@ func Navbar(currentPath string, links []PageLink) g.Node {
 				NavbarLink("/", "Home", currentPath == "/"),
 
 				// We can Map custom slices to Nodes
-				g.Group(g.Map(len(links), func(i int) g.Node {
-					return NavbarLink(links[i].Path, links[i].Name, currentPath == links[i].Path)
+				g.Group(g.Map(links, func(pl PageLink, _ int) g.Node {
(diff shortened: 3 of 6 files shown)


markuswustenberg:
Hey @codemicro. Thank you for the contribution!

I've also played around with a generics implementation recently, also for `Map` (pretty much the same you've done, but dropping the index int) but also trying to figure out if there are other places where it makes sense to introduce type parameters. So far, I haven't found anything other than `Map`, so I've been holding off on making this change for the same reasons you mention: it's a breaking change, and it requires Go 1.18.

For those reasons, I don't think I want to introduce generics into the library just yet.

codemicro:
That's perfectly understandable! This was just a shot in the dark, since I had a few spare minutes and felt like playing with it.

Thanks for taking the time to look over it, though :)

markuswustenberg:
BTW, I'm trying to figure out if type parameters could be a nice way to make a distinction between HTML elements, attributes, and other kinds of DOM nodes in the type system, while pretty much keeping the current API. So far, I haven't come up with anything good without using hacky things like tagging interfaces.

markuswustenberg:
> That's perfectly understandable! This was just a shot in the dark, since I had a few spare minutes and felt like playing with it.
> 
> Thanks for taking the time to look over it, though :)

It's really appreciated! 😊

codemicro:
> BTW, I'm trying to figure out if type parameters could be a nice way to make a distinction between HTML elements, attributes, and other kinds of DOM nodes in the type system, while pretty much keeping the current API. So far, I haven't come up with anything good without using hacky things like tagging interfaces.

This is a really interesting idea, though I'll admit I also don't really have any good ideas for it either. I'm still quite new to generics in Go myself, I've only written a little bit of code using it thus far.

codemicro:
(Also, I just wanna note that I'm pretty interested in this project - I've used [github.com/stevelacy/daz](https://github.com/stevelacy/daz) a decent bit in the past, but this seems like a more mature and usable version of that.)

markuswustenberg:
> (Also, I just wanna note that I'm pretty interested in this project - I've used [github.com/stevelacy/daz](https://github.com/stevelacy/daz) a decent bit in the past, but this seems like a more mature and usable version of that.)

That's nice to hear! I didn't know about daz, looks pretty similar in some respects.

I've seen quite a few projects that do something similar to gomponents. I think the main distinction comes down to the API (basically look and feel of writing components) and simplicity. I'm really trying hard to create something that looks like HTML, but is actually Go, while keeping it as simple as possible.

If you're interested, I could tag you here on Github if I ever come up with an API design I could use some feedback on? (I recently did that with others in #101, for example.) Then you can always decide whether you want to answer or not, obviously. 😊

codemicro:
Please do!

PR #137, opened 2023-05-11
Add MapIndex function

markuswustenberg:
I've often found myself needing the index when mapping something to nodes, so here's a `MapIndex` function.

### DIFF
--- a/gomponents_generic.go
+++ b/gomponents_generic.go
@@ -11,3 +11,12 @@ func Map[T any](ts []T, cb func(T) Node) []Node {
 	}
 	return nodes
 }
+
+// MapIndex is like Map, but also passes the index to the callback function.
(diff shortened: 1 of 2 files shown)


gomponents_generic.go
```diff
@@ -11,3 +11,12 @@ func Map[T any](ts []T, cb func(T) Node) []Node {
 	}
 	return nodes
 }
+
+// MapIndex is like Map, but also passes the index to the callback function.
+func MapIndex[T any](ts []T, cb func(T, int) Node) []Node {
+	var nodes []Node
```
the-goodies, line 17: nodes := make([]Node, 0, len(ts))

markuswustenberg:
I've decided not to do this, to keep the API surface area low.

PR #142, opened 2023-07-16
Added MapKey helper

demetrius-oak:
I needed a MapKey function, similar to Map, that maps a map[string]any

### DIFF
--- a/gomponents_generic.go
+++ b/gomponents_generic.go
@@ -11,3 +11,12 @@ func Map[T any](ts []T, cb func(T) Node) []Node {
 	}
 	return nodes
 }
+
+// Map a map[key:string]any to a slice of Nodes.
+func MapKey[T any](ts map[string]T, cb func(string, T) Node) []Node {
+	var nodes []Node
+	for k, t := range ts {
+		nodes = append(nodes, cb(k, t))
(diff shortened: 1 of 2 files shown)


markuswustenberg:
Hi @demetrius-oak, thank you for taking the time to contribute. 😊 I’m currently on paternity leave and will have a look at this with some delay.

gomponents_generic.go
```diff
@@ -11,3 +11,12 @@ func Map[T any](ts []T, cb func(T) Node) []Node {
 	}
 	return nodes
 }
+
+// Map a map[key:string]any to a slice of Nodes.
+func MapKey[T any](ts map[string]T, cb func(string, T) Node) []Node {
+	var nodes []Node
```
cezarguimaraes, line 17:
```suggestion
	nodes := make([]Node, 0, len(ts))
```

markuswustenberg:
Hi @demetrius-oak . Thank you for your patience!

Since maps in Go don't have a deterministic sorting order, this would produce a different result on every render. Therefore, I would rather leave it up to the user of the library to convert the map to a slice of a predetermined order, and then use the regular `Map` instead.

PR #144, opened 2023-07-29
Support HTMX and Hyperscript

blufor:
This allows building interactive web pages using [HTMX](https://htmx.org/) and [Hyperscript](https://hyperscript.org/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Made in 🇩🇰 by [maragu](https://www.maragu.dk), maker of [online Go courses
   - Type safety
   - Auto-completion
   - Nice formatting with `gofmt`
+- Use [HTMX](https://htmx.org/) and [Hyperscript](https://hyperscript.org/) for building interactive web applications
 - Simple API that's easy to learn and use (you know most already if you know HTML)
 - Useful helpers like `Text` and `Textf` that insert HTML-escaped text, `Map` for mapping data to components,
   and `If` for conditional rendering.
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
-module "github.com/maragudk/gomponents"
+module github.com/maragudk/gomponents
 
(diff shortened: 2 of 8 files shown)


markuswustenberg:
Hi @blufor , and thank you for contributing! Unfortunately what you’re looking for is already in the https://github.com/maragudk/gomponents-htmx repo. I haven’t looked at your code as I’m currently on paternity leave, but I think it’ll satisfy your needs? 😊

blufor:
@markuswustenberg ouch. well, that was quite a waste of time then.

However I see some of the HTMX supported attributes missing - those not having `hx-` prefix, like `sse-*` and `ws-*` and also the underscore (`_`) attribute that HTMX's buddy-scripting-language Hyperscript leverages. I'll make a PR into the repo you referred.

Additionally, It would be nice to mention the htmx-specific repo on gomponents.com. Or maybe making it more visible. That would definitely stop me from writing a whole new implementation :D Besides that, HTMX deserves to spread ;)

Thank you

markuswustenberg:
@blufor yeah unfortunately it probably was. I know it doesn't help much now, but another time it would be a good idea to open an issue discussing a larger change first, as also mentioned in [CONTRIBUTING.md](https://github.com/maragudk/gomponents/blob/main/CONTRIBUTING.md).

I think the `ws-` and `sse-` prefixes are for plugins? I haven't had a look at that part. I'm currently on paternity leave, so it'll probably be a while before I will look at any code. I'm not sure I want to integrate anything that's not core to HTMX itself in the library, but we'll see. Of course, you're free to make a fork and do whatever changes you want. 😊

By the way, other gomponents-adjacent libraries are mentioned at https://www.gomponents.com/plus/

Thanks again for taking the time to trying to improve the core library, even though it didn't work out this time. 😊

Issue #152, opened 2023-10-25
Inline if with nil pointer


onordgren:
Hello! Ran into the following issue today while trying gomponents. When using the inline if, `g.If` it seems like the node is evaluated even though the statement is false. Made a small example app below, where session `s` is `nil`. There is an if statement checking if the session is not nil, and then proceeding to show the username. The code however panics with a nil reference error. Looking at the source code it would seem like the node wouldn't be evaluated, but maybe something else is going on?

```
package main

import (
	"net/http"

	g "github.com/maragudk/gomponents"
	h "github.com/maragudk/gomponents/html"
)

type session struct {
	Username string
}

func main() {
	s := getSession()

	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		g.Node(
			h.Div(g.Text("Hello!"),
				g.If(s != nil,
					h.Div(g.Text(s.Username),
						g.Text("You are logged in!"),
					),
				),
			),
		).Render(w)
	})

	http.ListenAndServe(":8080", nil)
}

func getSession() *session {
	return nil
}
```

markuswustenberg:
Hi @onordgren , thank you for taking the time to file an issue! My answer time is a bit longer than usual, I've got the flu. 🤧

I agree that this is surprising behaviour, but actually a consequence of the design of gomponents itself. As you probably know, a component is basically (most often) a function. The functions are called inline in your example, and the component given to the `g.If` is always evaluated, but the result may be discarded. That's the reason for your nil pointer error.

I've explored possible solutions for this in #99 and #101, but haven't found a good one yet. Maybe I should use this opportunity to revisit the problem. Feedback and input appreciated. 😊

markuswustenberg:
I tried an implementation for a solution, but I'm not entirely satisfied. If anyone wants to have a look and post their thoughts, they are very welcome!

egeozcan:
Hi, a new user of the library here!

My humble suggestion is deprecating g.If and suggesting creating a function with an early nil return if the parameter does not satisfy. The language simply does not have the extensibility to enable any more elegant solution. But of course, I may be wrong and any pointers to alternative approaches would help a lot.

egeozcan:
hi @markuswustenberg, well I meant not using a generic function, and using a custom function to do this. the g.node is super easy to compose so I humbly suggested that, that should be encouraged instead of offering abstractions that can't cover some cases such as the one discussed and may cause confusion. thanks for the detailed reply 😊

[...]

PR #153, opened 2023-10-31
Add Lazy implementation

markuswustenberg:
See #152


markuswustenberg:
This doesn't actually work… I don't know what I was thinking.

PR #156, opened 2023-12-12
Add Lazy function

### LINKED ISSUE #152: Inline if with nil pointer

markuswustenberg:
`Lazy` provides lazy evaluation of node rendering. Useful together with `If`.

Fixes #152

### DIFF
--- a/gomponents.go
+++ b/gomponents.go
@@ -254,6 +254,9 @@ func Group(children []Node) Node {
 
 // If condition is true, return the given Node. Otherwise, return nil.
 // This helper function is good for inlining elements conditionally.
+// This doesn't behave entirely like a regular if statement, because nested Nodes and their parameters are evaluated
+// from the deepest level and up before hitting If (because that's how Go works).
+// If you need lazy evaluation of the given Node, see Lazy.
 func If(condition bool, n Node) Node {
 	if condition {
 		return n
@@ -264,7 +267,7 @@ func If(condition bool, n Node) Node {
 // Lazy returns a Node which calls the given callback function only when rendered.
 // Used with If, it enables lazy evaluation of the node to return, depending on the condition passed to If.
 // It defaults to being an ElementType, but can be set to AttributeType by passing a NodeType.
-// Even though nodeTypes is variadic, only the first NodeType is used.
+// Even though nodeType is variadic, only the first NodeType is used.
 func Lazy(cb func() Node, nodeType ...NodeType) Node {
 	if cb == nil {
 		panic("lazy callback cannot be nil")
--- a/gomponents.go
+++ b/gomponents.go
@@ -295,10 +295,3 @@ func (l *lazy) Render(w io.Writer) error {
 func (l *lazy) Type() NodeType {
 	return l.T
 }
-
-// String satisfies fmt.Stringer.
-func (l *lazy) String() string {
-	var b strings.Builder
-	_ = l.Render(&b)
-	return b.String()
-}
--- a/gomponents.go
+++ b/gomponents.go
@@ -260,3 +260,45 @@ func If(condition bool, n Node) Node {
 	}
 	return nil
 }
+
+// Lazy returns a Node which calls the given callback function only when rendered.
+// Used with If, it enables lazy evaluation of the node to return, depending on the condition passed to If.
+// It defaults to being an ElementType, but can be set to AttributeType by passing a NodeType.
+// Even though nodeTypes is variadic, only the first NodeType is used.
+func Lazy(cb func() Node, nodeType ...NodeType) Node {
+	if cb == nil {
+		panic("lazy callback cannot be nil")
+	}
+
+	switch len(nodeType) {
+	case 0:
+		return &lazy{CB: cb}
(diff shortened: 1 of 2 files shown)


egeozcan:
Hi, thanks for asking for feedback.

I like the idea. It's simple (both usage and the implementation), it works. I looked in detail and can't find anything to improve, implementation-wise. 

Perhaps docs need to be expanded (with a caution to prefer this when using If). If it weren't a breaking change, I'd even suggest to allow only lazy functions as the second argument to If, but maybe too much of a hassle for medium+ projects using the lib at this point.

markuswustenberg:
@egeozcan thank you for the quick feedback!

Good idea with the docs, I'll add that to `If` as well.

I considered a breaking change, but since I'd hate to break existing consumers and `If` works well for simple cases, I've chosen to keep it. Plus, `Lazy` can be used in other places where lazy evaluation is needed, for whatever reason. (I explored an `IfFunc` in #99.)

The only thing I'm not entirely happy with is the variadic `NodeType` to support attributes. It's kind of a hack, but I can't think of anything better. Alternatives considered:

- Adding an option struct to the callback. Discarded because: Seems messy for the default case (element type)
- Making a struct instead of a function. Discarded because: More verbose. This is basically #153.

I'll leave this open for a while to see if anyone has feedback.

markuswustenberg:
@egeozcan thank you for your response. Yeah, I considered two functions as well, but I think I'll go with my current one even though it's a bit more "magic" (in a bad way). Maybe I'll think of something better later.

Thank you again for your input!

hastebrot:
I wrote this today:

```go
func Iff(condition bool, node func() g.Node) g.Node {
	if condition {
		return node()
	}
	return nil
}
```

Usage

```
Iff(props.subtitle != nil, func() g.Node {
	return Span(g.Text(*props.subtitle))
}),
```

I thought `If()` might be a problem, since it builds the document tree for all nodes, even when the condition fails, as far as I understand. only the `Render()` stops at the node.

but `Lazy()` even looks better than my attempt:

```go
g.If(u != nil, g.Lazy(func() g.Node {
	return g.El("span", g.Text(u.Name))
})),
```

markuswustenberg:
@hastebrot yeah, your Iff is maybe how I should have implemented it in the first place. But in the end, I guess both have their usage, and I'm hoping I've found a sweet spot with Lazy. 😊

gomponents.go
```diff
@@ -254,9 +254,47 @@ func Group(children []Node) Node {
...
 		return n
 	}
 	return nil
 }
+
+// Lazy returns a Node which calls the given callback function only when rendered.
+// Used with If, it enables lazy evaluation of the node to return, depending on the condition passed to If.
+// It defaults to being an ElementType, but can be set to AttributeType by passing a NodeType.
+// Even though nodeType is variadic, only the first NodeType is used.
+func Lazy(cb func() Node, nodeType ...NodeType) Node {
```
vikstrous2, line 271: Two thoughts:
1. This seems to be complicated by implementing the optional nodeTypeDescriber interface, but I've been reading through the code and it seems like the only use of nodeTypeDescriber is to omit the output if the type is wrong here https://github.com/maragudk/gomponents/blob/1f0bb2a2aaecc7813b9eda1c0579434e9ad5fe56/gomponents.go#L110 so maybe Lazy can just not implement it without much being lost
2. If NodeType is important, and there are only two node types, maybe create a LazyAttr function and a LazyEl function?

markuswustenberg: @vikstrous2 thank you for the review! Sorry for the delay in replying, the winter and its various sicknesses has not been kind to me and my family this year… 🤒

Yes, `nodeTypeDescriber` is only used for checking whether a node should be rendered as an element or attribute. I think it's pretty useful in e.g. form checkboxes and the `checked` attribute, and it feels wrong to leave out support for attributes, even though it would simplify the API quite a bit here.

`LazyAttr` would also solve this, arguably more elegantly, but I would really like there to not be two functions for this already slightly edge-case functionality.

Hmm. I'm seriously considering dropping this whole `Lazy` stuff, it doesn't quite feel right.

markuswustenberg:
See #172 as a solution instead.

PR #160, opened 2024-01-15
Allow for groups to be rendered

Hendrikto:
Hey Markus,

What do you think about allowing groups to be rendered? I ran into this problem several times, while doing stuff like this:

```go
[…]
Td(g.Group(g.Map(subsidiaries, func(subsidiary models.Entity) g.Node {
	return g.Group([]g.Node{
		A(Href(fmt.Sprintf("/entity/%d", subsidiary.Entity_ID)),
			g.Textf("%d: %s (%s)", subsidiary.Entity_ID, subsidiary.Name, subsidiary.Country),
		),
		Br(),
	})
})))
[…]
```

Also, this limitation often stands in the way of straightforwardly extracting components into their own functions. You always have to have a wrapper, even when it does not make much sense.

It is pretty easy to add group rendering support, so maybe there is something I missed. What was the rationale for not allowing it in the first place? Are there deeper implications of this change I missed?

### DIFF
--- a/gomponents.go
+++ b/gomponents.go
@@ -237,12 +237,21 @@ type group struct {
 
 // String satisfies fmt.Stringer.
 func (g group) String() string {
-	panic("cannot render group directly")
+	var b strings.Builder
+	_ = g.Render(&b)
+	return b.String()
 }
 
 // Render satisfies Node.
-func (g group) Render(io.Writer) error {
-	panic("cannot render group directly")
+func (g group) Render(w io.Writer) error {
+	for _, node := range g.children {
+		err := node.Render(w)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 // Group multiple Nodes into one Node. Useful for concatenation of Nodes in variadic functions.


markuswustenberg:
Hi @Hendrikto, thank you for your contribution! It basically has to do with how components are rendered in `El`. It's a two-pass process, where attribute children are rendered first, then child elements. 

If this restriction weren't there, it would be possible to render nonsensical output such as ` style="" <span></span>`, which I chose to disallow from the beginning to make it harder to use wrong. 

That said, you're not the first to be confused by this, and I'm not sure I've hit the right solution. This is still something I'm considering before hitting v1.

Hendrikto:
Thanks for the quick reply and explanation. I had the feeling that I was missing something.

A few questions: Does that mean this PR should be closed?

How would you go about representing something like this?

```go
func entityLinks(entities []models.Entity) g.Node {
	return g.Group(g.Map(entities, func(entity models.Entity) g.Node {
		return g.Group([]g.Node{
			entityLink(entity),
			Br(),
		})
	}))
}
```

I could, of course implement my own `Concat` function, which would do what my implementation of `g.Group` did. But maybe I am just approaching the problem from the wrong angle.

The inner `g.Group` call is only needed because of the type of `g.Map`. What do you think about implementing something like `g.FlatMap`?

The outer `g.Group` is needed, because of the type of `g.If`. I use it as

```go
g.If(len(subsidiaries) > 0, entityLinks(subsidiaries))
```

Maybe `g.If` could be made variadic? But I guess doing that the naive way would also cause the same problems as allowing direct `g.Group` rendering.

Putting both ideas together, I could do something like

```go
func entityLinks(entities []models.Entity) g.Node {
	return g.FlatMap(entities, func(entity models.Entity) []g.Node {
		return []g.Node{
			entityLink(entity),
			Br(),
		}
	})
}
```

and then use it like

```go
g.If(len(subsidiaries) > 0, entityLinks(subsidiaries)...)
```

What do you think?

markuswustenberg:
Yeah, I don't think I'll merge this PR as-is, I need to figure out what approach to take with `Group`. You can of course always just use a small, private helper function to do the same thing in your code base.

Regarding your `entityLinks`, what problems are you seeing with the approach you're already using? The `Group` call should result in a `Node` that renders just fine when rendered through a parent element, so your use of `If` should also work as expected. Whether you return `g.Node` or `[]g.Node` is more a choice of style, both would work fine depending on the context.

Hendrikto:
> Yeah, I don't think I'll merge this PR as-is, I need to figure out what approach to take with Group.

Yeah, after some more consideration, I think both approaches I suggested above would actually result in the same problems as allowing direct group rendering.

> Regarding your entityLinks, what problems are you seeing with the approach you're already using? The Group call should result in a Node that renders just fine when rendered through a parent element, so your use of If should also work as expected.

As you said, it works fine when rendering through another element. The problem is that I am using HTMX, and I want to be able to render the `entityLinks` component on its own, without a parent element. There are ways of working around this limitation, but they all come with their own drawbacks, and are not very clean.

markuswustenberg:
> As you said, it works fine when rendering through another element. The problem is that I am using HTMX, and I want to be able to render the `entityLinks` component on its own, without a parent element. There are ways of working around this limitation, but they all come with their own drawbacks, and are not very clean.

Ah, right. Thank you for that use case, I'll keep that in mind. 

Have you seen the outerHTML swapping method in HTMX? You could use that to replace the whole element including the parent, maybe that works for you. https://htmx.org/docs/#swapping

Otherwise, I'd say the best approach is to just make a custom `Group` function and use that for now. 

By the way, check out [gomponents-htmx](https://github.com/maragudk/gomponents-htmx) if you haven't already. :)

Hendrikto:
> Have you seen the outerHTML swapping method in HTMX?

Yes, but that is also a bit problematic, when trying to use the same component in different contexts. Sometimes the parent element is a `td`, sometimes a `nav`.

You can combine it with `hx-select`, but that is also not ideal. In general, I conceptually just don't like having to put a parent element just for the sake of rendering, and then stripping it back out again.

> Otherwise, I'd say the best approach is to just make a custom Group function and use that for now.

Thanks for the advice :)

> By the way, check out [gomponents-htmx](https://github.com/maragudk/gomponents-htmx) if you haven't already. :)

Yeah, I am already using it, but ran into a small problem there as well. However, this outside of the scope of this PR. Should I open an issue on that repo?

I am closing this PR for now. Thanks for your responsiveness and helpfulness :+1:

markuswustenberg:
@Hendrikto always feel free to open issues! 😊 Please do.

You're very welcome.

PR #176, opened 2024-06-18
Add dir

amrojjeh:
`dir` is a global attribute that lets you specify whether text is RTL or LTR or if it should be determined by the browser.

Documentation: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir


markuswustenberg:
Hey @amrojjeh , thank you for the contribution! 😊

PR #177, opened 2024-06-20
Add datetime attribute

JulienTant:
`<time>`, `<del>` and `<ins>` can receive a [`datetime`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTimeElement/dateTime) attribute


markuswustenberg:
@JulienTant thanks!

Issue #167, opened 2024-02-22
HTML -> gomponents converter tool


erikd234:
I really think this is a cool project and I am giving it a try.

Any advice for converting my current HTML files to gomponents? Either a tool to convert html files to gomponents, or allowing html to mix with gomponents so I can render html then some gomponents within my web page

IE
```
<html>
<body>
  <h1>Outer shell<>
  <div> <--  HERE INSERT GOCOMPONENTS-->
```  
  
  I just have alot of html and would love to prototype designs in html then convert them to gomponents afterwards.
  
  What do you think of this idea?

markuswustenberg:
Hi @erikd234 . I've thought about doing a converter from HTML to gomponents, but haven't tried it yet. I think that could be an interesting little project!

What I usually do when I am in this situation is to put the relevant HTML in files next to your .go component files, embed them with `go:embed`, and then use `g.Raw` to render them directly. The resulting overall component sometimes looks a bit odd, but it works just fine. Then you can selectively replace HTML with gomponents where needed.

Would that work for you?

markuswustenberg:
Note to self: there's an HTML parser: https://pkg.go.dev/golang.org/x/net/html

PiotrKowalski:
Hi.

I started using Gomponents with Tailwind and I started hating rewriting HTML classes into Gomponents code.
So I wrote a small helper tool. Also, I plan to add exposed api for sending files to server and returning responses via endpoint. Later I will also add CLI tool

http://htg.piotrkowalski.me/ 
It runs on my local computer so may not be 24/7, feel free to use it. I dont handle most of gomponents elements and attributes now. Right now, I mostly take the html node data and uppercase the first letter. The custom HTML to gomponents can be easily added. WIll do it if I come accross some problem myself.

You can check source code here 
https://github.com/PiotrKowalski/html-to-gomponents

markuswustenberg:
@PiotrKowalski , that is so super cool! :D Could you host it somewhere that can take some traffic? Then I'll link to it from the gomponents website.

PiotrKowalski:
@markuswustenberg 
Thanks for the acknowledgment

I will look into it, but for ~one week I wont be able to do anything more in this matter.
And also, it is like 0.0.1 version. I need to add more gomponents structs there. For example, I am not able to create g.Attr() node automatically right now.
This converter handles like 80% of manual labour, remaining 20% needs to be done manually by dev for now.

WIll look into hosting next week

[...]

PR #178, opened 2024-06-23
Convert HTML to Gomponents

amrojjeh:
This is my attempt to solve #167. I didn't use [x/net/html](https://pkg.go.dev/golang.org/x/net/html) because... well I forgot to look into it. The XML parser in the standard library seems to work fine though once configured to parse HTML.

The way I decided to format the output was as such:
- Attributes are on the same line
- `g.Text` and other elements take up a new line
- Whitespace is ignored
- Tabs are used instead of spaces (as per Go's standard)
- It's standard to have one "page" per file, so this code outputs an entire new file with the imports and the package (default "ui") rather than outputting a single function

Here's an example:
```html
<!DOCTYPE html>
<html lang="en">
<head>
	<meta charset="UTF-8">
	<meta name="viewport" content="width=device-width, initial-scale=1.0">
	<title>Document</title>
</head>
<body>
	<p>How are you doing?</p>
	<unknown required data-crazy="lol">wow</unknown>
	<a aria-hidden="true" data-wow="some field" unrecognized="crazy" anotherunrecognized></a>
	<form action="/test" method="get">
		<input type="text" required>
	</form>
</body>
</html>
```

```go
package ui

import (
	. "github.com/maragudk/gomponents/html"
)

func Page() g.Node {
	return Doctype(
		HTML(Lang("en"),
			Head(
				Meta(Charset("UTF-8")),
				Meta(Name("viewport"), Content("width=device-width, initial-scale=1.0")),
				TitleEl(
					g.Text("Document"),
				),
			),
			Body(
				P(
					g.Text("How are you doing?"),
				),
				g.El("unknown", Required(), Data("crazy", "lol"),
					g.Text("wow"),
				),
				A(Aria("hidden", "true"), Data("wow", "some field"), g.Attr("unrecognized", "crazy"), g.Attr("anotherunrecognized", "anotherunrecognized")),
				Form(Action("/test"), Method("get"),
					Input(Type("text"), Required()),
				),
			),
		),
	)
}
```
(**Note:** the output contains trailing commas, not seen here, but these are easily removed by running `go fmt`, which is integrated into most editors)

Usage: `go run . (-p PACKAGE NAME) [html file]`

There is still lots of work left, such as integrating SVG and HTMX, running automated tests, supporting globbing, but I think this lays a good foundation.

Apologize for the older commits as well. I had an older fork of this repo and I didn't know how to squash them excerpt by deleting my fork and starting a new one, and at that point I had already started working on it.

Feel free to move this off of main.

### DIFF
--- a/converter/main.go
+++ b/converter/main.go
@@ -203,7 +203,7 @@ func main() {
 
 func convert(r io.Reader, w io.Writer, packageName string) {
 
-	w.Write([]byte(fmt.Sprintf(`package %s
+	write(w, fmt.Sprintf(`package %s
 
 import (
 	g "github.com/maragudk/gomponents"
@@ -211,7 +211,7 @@ import (
 )
 
 func Page() g.Node {
-	return `, packageName)))
+	return `, packageName))
 
 	dec := xml.NewDecoder(r)
 	dec.Strict = false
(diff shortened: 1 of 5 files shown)


amrojjeh:
Tests are failing because I accidentally used go 1.20 features :skull: ([`strings.CutPrefix`](https://pkg.go.dev/strings#CutPrefix))

markuswustenberg:
Hey @amrojjeh, thanks for having a go at this! I'll see if I can get around to having a look at it before the start of my summer holidays at the end of this week. 😊

markuswustenberg:
Hi @amrojjeh . Thanks again for having a go at this!

Since then, @PiotrKowalski made an online version at https://htg.piotrkowalski.me, with source available at https://github.com/PiotrKowalski/html-to-gomponents . I think I'm going with pointing to that for now, since it's the least amount of work for me and looks like it's working fine. If I'm adding anything to the core library, I'd have to make sure to understand it and be responsible for maintaining it, and unfortunately I don't want to take that on right now.

But perhaps it would make sense for you to publish this as an independent CLI tool?

PR #185, opened 2024-07-04
Update attributes.go - add draggable attribute

sombriks:
add draggable attribute


markuswustenberg:
Thanks! Could you add it to the table test as well?

markuswustenberg:
Oh, could you remove the changes to .gitignore as well? :)

PR #188, opened 2024-07-28
Add new `Hreff` attribute function

gabrielhora:
Hey there, I've written this enough times and I feel like it makes sense to have that quality of life function. Do you think that it would be useful to add it to the library itself to save a few key strokes?

---

Hreff allows formatting the attribute value with arguments, like `fmt`. So instead of writing:

```golang
Href(fmt.Sprintf("/some-path/%d", obj.ID))
```

You can directly format the path:

```golang
Hreff("/some-path/%d", obj.ID)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /cover.out
+.idea
--- a/html/attributes.go
+++ b/html/attributes.go
@@ -1,6 +1,8 @@
 package html
 
 import (
+	"fmt"
+
 	g "github.com/maragudk/gomponents"
 )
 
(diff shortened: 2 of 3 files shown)


markuswustenberg:
Hi @gabrielhora . Thank you for taking the time to send in a suggestion! 😊 It's really appreciated.

Unfortunately, I'd like to keep the `html` package purely for elements and attributes that exist in the specs, no helpers. I would suggest you maybe make your own helper module you always import that you can use across projects.

Good luck with your gomponents projects!

PR #196, opened 2024-08-30
Add SVG Attributes and Elements

tevintchuinkam:
Added some missing svg attributes and elements by looking at the [mozilla docs](https://developer.mozilla.org/en-US/docs/Web/SVG). 

Tests and readme where updated accordingly.


markuswustenberg:
Hi @tevintchuinkam. Thank you for the PR! I'll have a look through it when I've got some spare time. 😊

markuswustenberg:
@tevintchuinkam, I've been thinking about this. I'm a but reluctant to add all this to the core gomponents library, as I don't know so much about SVG myself, and I'm wondering whether it can give problems down the road to include it in an HTML rendering library.

For example, this will break code where you dot-import both HTML and SVG elements, because SVG has its own `style`.

I've been thinking about creating a gomponents-svg library instead, focused on rendering SVG/XML. I think it may have been a mistake to include the few SVG items I have already.

Thoughts?

tevintchuinkam:
@markuswustenberg  I didn't consider that. You're right.

PR #202, opened 2024-09-19
Make `Group` a type

### LINKED ISSUE #201: Add varargs to Group?
In addition to the slice of Nodes.

markuswustenberg:
This changes `Group` to be a type instead of a function, which means it can support both grouping `[]Node` as well as doing variadic-ish things like `Group{n1, n2}`.

This also means that `Map` can just return a `Group`.

Special thanks to @deitrix for this simple and elegant solution that, for some reason or another, has had me perplexed for a long time.

Fixes #201.


markuswustenberg:
Not sure yet if this is a good idea or confusing, leaving it open for a while.

EwenQuim:
Why not simply use the variadic syntax?

I know it would break backward compatibility and force people to write things like

```go
g.Group(g.Map(l, f)...)
// Instead of
g.Group(g.Map(l, f))
```

But that's ok I guess, and the error is screaming (not a silent breaking change)

markuswustenberg:
@EwenQuim thanks for having a look!

Basically because I don't want to break that particular style of usage, with `Map` inside `Group`. It's so common everywhere, and I use it myself, and I'd rather not force people to use the spread syntax.

I may just drop this PR and accept that, in the particular use case of returning a list of `Node`s without a parent, that one will have to write

```go
g.Group([]g.Node{
…
})
```

Still thinking about this one. :D

deitrix:
For cases like `Map`, might it make sense to return a slice type that implements `Node`, which would effectively be like passing the result to `Group`?

```go
type NodeList []Node

func (nl NodeList) Render(w io.Writer) error {
	return Group([]Node(nl)).Render(w)
}
```

```go
func Map[T any](ts []T, cb func(T) Node) NodeList
```

deitrix:
Alternatively:
```go
type Group []Node
```
```go
func Map[T any](ts []T, cb func(T) Node) Group
```

EwenQuim:
I looove this idea. Really Go-esque and elegant solution

deitrix:
@markuswustenberg It would be breaking change, but most standard uses should _just work_. `Group` as a slice type would be a drop-in replacement for callers of the `Group` function. The only case I can think of where it would throw a compiler error is where the caller expected `Group` to be a function type, though I can't see that kind of usage being very common.

`Group(Map(...))` will also continue to work, with the `Group(...)` being superfluous if `Map`'s signature is updated to return a `Group`

I tested the change on my (albeit rather small) project, and didn't get any compiler errors! Might be worth testing on some larger projects, if you have any to hand?

deitrix:
> The only case I can think of where it would throw a compiler error is where the caller expected Group to be a function type, though I can't see that kind of usage being very common.

With this in mind, your original PR would also be considered a breaking change 😁

markuswustenberg:
> Alternatively:
> 
> ```go
> type Group []Node
> ```
> 
>  ```go
> func Map[T any](ts []T, cb func(T) Node) Group
> ```
> 
> Then, you have the option between passing a slice (`Group(children)`), or varargs (`Group{child1, child2, ...}`).

@deitrix you're a genius! I didn't understand the full implications of this earlier today. It's so simple and elegant, I can't believe I haven't thought of this solution before. _Of course_ a `Group` is just a list of nodes. 😅

All tests pass, this looks great. I'm going to leave it open for a few days to ponder implications and test it on my codebases, but I think it's good.

markuswustenberg:
This has been nagging me for literally _months_!

markuswustenberg:
@eduardolat @JulienTant since you were both involved in the `Fragment` discussions around #181, I think you'll be pleased to hear that this looks like a final, elegant API for `Group` that basically solves All The Things.

eduardolat:
Looks great, thanks for your work on this fantastic project

markuswustenberg:
I'm just going to merge this, I think it's great. Thanks again, @deitrix, @EwenQuim, and @eduardolat.

daaku:
This is a super nice change! Thank you all 👍

EwenQuim:
Also, note that Group must now (in most cases) be used directly as a type, as Go types that are hidden slices can be directly used as it is

image



image

markuswustenberg:
@EwenQuim I think it takes some adjusting and documentation, but otherwise I think the `Group{…}` style is actually quite neat, and what you would expect from a slice of something. It also functions much like the `Classes` helper (except that one's a map, not a slice). So I don't think I'll add a `Fragment` helper, I would consider it redundant.

Regarding your last example: you could just put different kinds of `Node`s into the slice, as long as they're satisfying the interface. So `Div(ID(""), Group{Span(), Div()}, Span())` would work as expected.

EwenQuim:
I think you're right, it's OK!

> Regarding your last example

Yeah that was not a problem, that was just a remark that slices arguments do not work like variadic args (variadic args are more flexible). Gomponents API is really well done :)

markuswustenberg:
@EwenQuim I'm glad we agree.

And thanks! I'm quite proud of it. 😁

PR #224, opened 2024-10-09
Add missing 'hidden' attribute

glsubri:
Hey @markuswustenberg 

I realised today that gomponents had no `hidden` attribute. So I thought I could add it.

However, quickly going through MDN, I realised that it could also take the 'until-found' string value (see [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden#the_hidden_until_found_state)).

I think it would be nicer to have something like this:

```go
func Hidden() g.Node {
	return g.Attr("hidden")
}

func HiddenUntilFound() g.Node {
	return g.Attr("hidden", "until-found")
}

func example() g.Node {
	return Div(
		Hidden(),
	)
}
```

Instead of 
```go
func Hidden(v string) g.Node {
	return g.Attr("hidden", v)
}

func example() g.Node {
	return Div(
		Hidden(""),
	)
}

```

I know that it deviates a bit from the way you've written the other attributes but this feels nicer to me.

What do you think ?


markuswustenberg:
@glsubri thanks for this!

HTML is just the gift that keeps on giving with new surprises. 😅

I value consistency _a lot_, so I would lean towards adding just:

```go
func Hidden(v string) g.Node {
	return g.Attr("hidden", v)
}
```

And letting the user figure it out according to the docs. Since `hidden`, `hidden=""`, and `hidden="hidden"` are equivalent, I wouldn't mind the additional empty string argument.

Thoughts?

glsubri:
Hey @markuswustenberg 

Works for me. I've made the changes 👍

markuswustenberg:
Thanks!

PR #228, opened 2024-10-12
Handle void tags with different render function

emad-elsaid:
# Context

* During rendering any tag `render` function will check if the tag name is void by looking it up from a map
* This means every tag has a small performance hit when rendered
* As the list of void tags is constant and doesn't change in runtime the decision of how to render can be taken in compile time

# Changes
* This PR removes the map of void tags
* Removes the condition to handle void tags in `render`
* Creates a new `VoidEl` function to create void tag node
* It creates a new `renderVoid` function which renders the tag without close tag.
* It changes the void tags functions to use `VoidEl` instead of `El` 
* And it adds a missing void tag functions that was in the map but didn't have a function (command, keygen, track)

Relevant [initial discussion](https://www.reddit.com/r/golang/comments/1g15g4a/comment/lrfz5kn/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button)

### DIFF
--- a/gomponents.go
+++ b/gomponents.go
@@ -71,11 +71,18 @@ func (n NodeFunc) String() string {
 // See https://dev.w3.org/html5/spec-LC/syntax.html#elements-0 for how elements are rendered.
 // No tags are ever omitted from normal tags, even though it's allowed for elements given at
 // https://dev.w3.org/html5/spec-LC/syntax.html#optional-tags
-// If an element is a void element, non-attribute children nodes are ignored.
 // Use this if no convenience creator exists in the html package.
 func El(name string, children ...Node) Node {
 	return NodeFunc(func(w io.Writer) error {
 		return render(w, &name, children...)
+	})
+}
+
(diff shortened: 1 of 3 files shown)


markuswustenberg:
Hi @emad-elsaid, thank you for wanting to contribute! 😊

Unfortunately this isn't a change I'd like in the library.

The reason is that I think you're basically moving the complexity from the library to the user, so they have to know whether an element is a void element or not (if they're not using the helpers from the HTML package). Compared to the tiny performance gain of not having to look up a key in a map, I don't think this is the right tradeoff.

I'm sorry that you spent time on code that isn't going to be accepted. For another time, feel free to open an issue first discussing the proposed change, as outlined in [CONTRIBUTING.md](https://github.com/maragudk/gomponents/blob/main/CONTRIBUTING.md).

emad-elsaid:
No worries, thanks for explaining the reasoning

PR #232, opened 2024-10-18
Add html2gomponents tool

markuswustenberg:
This is a separate tool for converting HTML to gomponents. I decided to have a go at this after all, like @amrojjeh got started in #178, but with the HTML parser built by the Go team. I actually like that there's an authorative converter tool.

This lives in an entirely separate module in the same repo. I'm not yet sure what the implications are of this, but I'm going to give it a try.

See #167


tevintchuinkam:
Nice that you are working on this, I need this all the time as well. 

Would be nice to have a command line tool for this eventually.

markuswustenberg:
Yeah @tevintchuinkam this is a command line tool. I'm thinking of putting it in a separate repo though, because the versioning for the library would conflict with any versioning for the tool, because versioning in Go can only use git tags AFAIK.

amrojjeh:
Cobra has their CLI in a separate repo so that would follow convention

traherom:
I continue to be lazy about wanting to move this to a separate repo, but here's the core conversion code that's powering [my conversion site](https://gomponents.morehart.dev). You could shove this into your CLI super easily, just call  `Convert(io.Stdin)`. Fancy people could even handle some flags and set the various options.

This doesn't do the go file package/function wrapper stuff because I didn't personally want that.

https://gist.github.com/traherom/cc4ae4661d457177ce256ac52e1c8637

PS, I love Gomponents.

markuswustenberg:
> PS, I love Gomponents.

❤️

PR #233, opened 2024-10-18
Add components.MergeClasses

markuswustenberg:
Inspired by https://www.reddit.com/r/golang/comments/1g5mkcm/best_htmlcssjs_component_library/lsi038g/

Not sure this will be merged yet, WIP.

### DIFF
--- a/components/components.go
+++ b/components/components.go
@@ -63,3 +63,22 @@ func (c Classes) String() string {
 	_ = c.Render(&b)
 	return b.String()
 }
+
+func MergeClasses(classes ...string) g.Node {
+	each := map[string]struct{}{}
+	for _, c := range classes {
+		parts := strings.Split(c, " ")
+		for _, p := range parts {
+			each[p] = struct{}{}
+		}
+	}
+	var result []string
+	for c := range each {
+		if c == "" {
+			continue
+		}
+		result = append(result, c)
+	}
+	sort.Strings(result)
+	return Class(strings.Join(result, " "))
+}


apetsko (changes requested):
Hi, just little faster implementation

	var result []string
	parts := strings.Split(strings.Join(classes, " "), " ")
	for _, p := range parts {
		if p == "" || slices.Contains(result, p) {
			continue
		}
		result = append(result, p)
	}
	sort.Strings(result)
	return strings.Join(result, " ")

pdf:
I'm undecided if this is a good idea or not yet, but on the implementation, do we care about duplicates? The current implementation for `Classes` doesn't dedup.

If we do want dedup, this implentation significantly more efficient:

```go
func MergeClasses(classes ...string) string {
        var results []string
        for _, c := range classes {
                results = append(results, strings.Split(c, " ")...)
        }

        sort.strings(results)
        return g.Node(strings.Join(slices.Compact(results), " "))
}
```

```
BenchmarkDefaultNonOverlapping-16         376383              2987 ns/op
BenchmarkDefaultOverlapping-16            361663              4005 ns/op
BenchmarkApetskoNonOverlapping-16         600678              1877 ns/op
BenchmarkApetskoOverlapping-16            536390              2237 ns/op
BenchmarkSlicesNonOverlapping-16         1495579               815.8 ns/op
BenchmarkSlicesOverlapping-16             894409              1271 ns/op
```

If we don't care about duplicates, this func just becomes a simple append.

markuswustenberg:
Thanks to both of you, but I'm more concerned about the API design rather than the performance currently. :)

I don't think I'll keep this PR, since it's more of a utility function, not a component, and I don't know where I should put those yet.

PR #235, opened 2024-10-25
Add web component elements and slot attribute

knpwrs:
The slot attribute is a standard attribute for use with web components: https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_templates_and_slots


markuswustenberg:
Thanks! I don’t think we have the `template` and `slot` elements either?

knpwrs:
I was thinking about that and couldn't think of a situation where I (in my workflow) would server-render `<template>` or `<slot>`. I can add them though, others may have a use-case.

markuswustenberg:
D’oh, we run into a naming collision of course. Could you suffix each of `Slot` with `El` and `Attr`? And not add the non-suffixed version for now.

knpwrs:
Should be all set now.

markuswustenberg:
Thanks! Now just add them to the test tables and we’re set. 😊

knpwrs:
Tests should be all set now

markuswustenberg:
🥳

PR #236, opened 2024-10-26
Add popover attributes

knpwrs:
This PR adds popover attributes as detailed here: https://developer.mozilla.org/en-US/docs/Web/API/Popover_API/Using


markuswustenberg:
Hi @knpwrs, thanks for this.

This is the second attribute in a short time that allows zero or one argument(s) (`popover` is the same as `popover="auto"`). I'm wondering whether we should start adding a third form of attribute in the `html` package, with varargs values, similar to how `gomponents.Attr` itself works. And then panic if there's more than one value. Thoughts?

markuswustenberg:
@knpwrs I think in-place is fine. But it needs to go into a new test which:
1. Tests that it works like a boolean attribute
2. Tests that it works as a key-value attribute
3. Tests that it panics with more than one value given

Might as well start a new table-driven test, I think more attributes like these will show up.

Are you up for that? Otherwise let me know. 😊

knpwrs:
How does that look?

markuswustenberg:
Looks great! Should we add the `popovertargetaction` now we're at it? :D

knpwrs:
Done!

PR #241, opened 2025-02-02
Add formxxx attributes (for <input type="submit"> element)

slegrand45:
Hello,

A small patch to add the following attributes:

- formaction
- formenctype
- formmethod
- formnovalidate
- formtarget

These attributes are useful in `<input type="submit">` element: [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit#additional_attributes)


markuswustenberg:
Thanks! 😊

PR #242, opened 2025-02-05
Add 'download' and 'referrerpolicy' attributes for anchor <a> element

slegrand45:
Hello,

A patch to add [`download`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#download) and [`referrerpolicy`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#referrerpolicy) for the anchor `<a>` element.


markuswustenberg:
I'm learning so much about HTML attributes I didn't know about these days. :D Thanks, @slegrand45!

PR #243, opened 2025-03-02
Improve performance of `Map` and `Classes.Render` by preallocating slices

gabe565:
Hey! I just discovered this repo and I love being able to write templates while still having compile-time checks.

I was looking through the code and noticed slices aren't currently preallocated in `Map()` and `Classes.Render()`. Preallocating will improve performance since the resulting slice won't have to be grown dynamically during the append loop.

### DIFF
--- a/components/components.go
+++ b/components/components.go
@@ -44,7 +44,7 @@ type Classes map[string]bool
 
 // Render satisfies [g.Node].
 func (c Classes) Render(w io.Writer) error {
-	var included []string
+	included := make([]string, 0, len(c))
 	for c, include := range c {
 		if include {
 			included = append(included, c)
--- a/gomponents.go
+++ b/gomponents.go
@@ -250,7 +250,7 @@ func Rawf(format string, a ...interface{}) Node {
 
 // Map a slice of anything to a [Group] (which is just a slice of [Node]-s).
 func Map[T any](ts []T, cb func(T) Node) Group {
-	var nodes []Node
+	nodes := make([]Node, 0, len(ts))
 	for _, t := range ts {
 		nodes = append(nodes, cb(t))
 	}


markuswustenberg:
Hey @gabe565 . Thanks for contributing!

I've actually intentionally kept the other style because I think it's easier to read, and I don't think the change will have any real-world performance impact. (I'd be happy to be proven wrong on that with some data based on realistic benchmarks.)

That said, since I think it comes down to preference and you're not the first to mention this, I'm going to accept it anyway. :D

gabe565:
You're right that the impact is small in most cases, but preallocating slices avoids multiple allocations and copies as they grow dynamically. This reduces memory usage and GC pressure, which can add up on high-traffic servers where a lot of mapping happens per request. Either way, thanks for the merge!

I wrote a quick benchmark, and it seems to be ~40% faster with 100 entries:

<details>
  <summary>Source</summary>

```go
package main

import (
	"strconv"
	"testing"

	g "maragu.dev/gomponents"
)

const count = 100

func newInput(n int) []string {
	result := make([]string, 0, n)
	for i := range n {
		result = append(result, strconv.FormatInt(int64(i), 10))
	}
	return result
}

func MapNoPrealloc[T any](ts []T, cb func(T) g.Node) g.Group {
	var nodes []g.Node
	for _, t := range ts {
		nodes = append(nodes, cb(t))
	}
	return nodes
}

func BenchmarkMapNoPrealloc(b *testing.B) {
	input := newInput(count)
	for b.Loop() {
		MapNoPrealloc(input, func(s string) g.Node { return g.Raw(s) })
	}
}

func MapWithPrealloc[T any](ts []T, cb func(T) g.Node) g.Group {
	nodes := make([]g.Node, 0, len(ts))
	for _, t := range ts {
		nodes = append(nodes, cb(t))
	}
	return nodes
}

func BenchmarkMapWithPrealloc(b *testing.B) {
	input := newInput(count)
	for b.Loop() {
		MapWithPrealloc(input, func(s string) g.Node { return g.Raw(s) })
	}
}
```
</details>

```
goos: darwin
goarch: arm64
pkg: github.com/gabe565/test
cpu: Apple M1 Pro
BenchmarkMapNoPrealloc
BenchmarkMapNoPrealloc-10      	  422380	      2615 ns/op
BenchmarkMapWithPrealloc
BenchmarkMapWithPrealloc-10    	  587340	      1856 ns/op
PASS
```

markuswustenberg:
@gabe565, thanks for the benchmark!

Since the difference is about 800 nanoseconds per operation on your machine, I think the change will not be noticeable on the CPU front in the vast majority of real-world deployments. But it certainly doesn't hurt either, so I appreciate the change. 😊

But I was a bit curious, so I changed the benchmark to include larger `count`s and benchmarked memory as well:

<details>
  <summary>main_test.go</summary>

```go
package main

import (
	"fmt"
	"strconv"
	"testing"

	g "maragu.dev/gomponents"
)

func newInput(n int) []string {
	result := make([]string, 0, n)
	for i := range n {
		result = append(result, strconv.FormatInt(int64(i), 10))
	}
	return result
}

func MapNoPrealloc[T any](ts []T, cb func(T) g.Node) g.Group {
	var nodes []g.Node
	for _, t := range ts {
		nodes = append(nodes, cb(t))
	}
	return nodes
}

func BenchmarkMapNoPrealloc(b *testing.B) {
	counts := []int{10, 100, 1000, 10000}
	for _, count := range counts {
		b.Run(fmt.Sprint(count), func(b *testing.B) {
			input := newInput(count)
			for b.Loop() {
				MapNoPrealloc(input, func(s string) g.Node { return g.Raw(s) })
			}
		})
	}
}

func MapWithPrealloc[T any](ts []T, cb func(T) g.Node) g.Group {
	nodes := make([]g.Node, 0, len(ts))
	for _, t := range ts {
		nodes = append(nodes, cb(t))
	}
	return nodes
}

func BenchmarkMapWithPrealloc(b *testing.B) {
	counts := []int{10, 100, 1000, 10000}
	for _, count := range counts {
		b.Run(fmt.Sprint(count), func(b *testing.B) {
			input := newInput(count)
			for b.Loop() {
				MapWithPrealloc(input, func(s string) g.Node { return g.Raw(s) })
			}
		})
	}
}
```
</details>

```shell
$ go test -bench=. -benchmem
goos: darwin
goarch: arm64
pkg: app
cpu: Apple M3 Max
BenchmarkMapNoPrealloc/10-16       4397709       255 ns/op        736 B/op         15 allocs/op
BenchmarkMapNoPrealloc/100-16       643531      1816 ns/op       6864 B/op        108 allocs/op
BenchmarkMapNoPrealloc/1000-16       73420     16335 ns/op      59184 B/op       1011 allocs/op
BenchmarkMapNoPrealloc/10000-16       4718    279679 ns/op     905978 B/op      10018 allocs/op

BenchmarkMapWithPrealloc/10-16     7130872       168 ns/op        400 B/op         11 allocs/op
BenchmarkMapWithPrealloc/100-16     839346      1438 ns/op       4192 B/op        101 allocs/op
BenchmarkMapWithPrealloc/1000-16     84445     14259 ns/op      40384 B/op       1001 allocs/op
BenchmarkMapWithPrealloc/10000-16     7725    155771 ns/op     403840 B/op      10001 allocs/op
PASS
ok      app     9.801s
```

(I've changed the indentation of the output so numbers right-align.)

So it looks like memory usage is also better especially on 10000 elements, while numbers of allocs is roughly the same.

In summary: I'll keep the change, because it certainly doesn't hurt. :D Thanks again, and enjoy gomponents!

PR #249, opened 2025-03-19
add Scope attribute helper

alanpearce:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/th#scope


markuswustenberg:
Thanks! 😊

PR #250, opened 2025-03-19
add MapWithIndex and MapMap helpers

### LINKED ISSUE #240: Additional Map function that passes the current index to the callback function
Just wanted to say that I've really been enjoying using this library! Thanks for taking the time to create it and making it open source :)

I'm wondering if you'd be open to including an additional mapping function like [`Map`](https://github.com/maragudk/gomponents/blob/main/gomponents.go#L252) but that also passes the current index to the `callback` function?

The code below is what I've implement in my projects. I figured creating a new function would be the better option because modifying the existing `Map` function would cause a breaking change since the `callback` function signature would change. Please let me know your thoughts!
```go
func MapWithIndex[T any](collection []T, callback func(index int, item T) gomponents.Node) gomponents.Group {
	nodes := make([]gomponents.Node, 0, len(collection))
	for index, item := range collection {
		nodes = append(nodes, callback(index, item))
	}
	return nodes
}
```

I can submit a PR once your happy with the implementation :)

alanpearce:
I found myself needing to map over a set, then I saw #240, so I thought I'd implement both at once.

Fixes #240

### DIFF
--- a/gomponents.go
+++ b/gomponents.go
@@ -257,6 +257,26 @@ func Map[T any](ts []T, cb func(T) Node) Group {
 	return nodes
 }
 
+// Map a slice of anything to a [Group] (which is just a slice of [Node]-s).
+func MapWithIndex[T any](ts []T, cb func(int, T) Node) Group {
+	nodes := make([]Node, 0, len(ts))
+	for k, t := range ts {
+		nodes = append(nodes, cb(k, t))
+	}
+
+	return nodes
+}
+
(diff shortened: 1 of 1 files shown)


markuswustenberg:
@alanpearce thank you for wanting to contribute this. Unfortunately it's not something I want to add currently.

The `Map` function is from before generics in Go were a thing, and I probably wouldn't add it now, given that there are other libraries that essentially solve the same thing, and there's nothing HTML-specific about them.

But I appreciate that you want to contribute to improving the library, thank you for that! 😊

PR #253, opened 2025-04-12
Combine class attributes

makinori:
Multiple `Class()`'s doesn't concatenate together. When using something like Tailwind, you can't easily add more classes from the parent element. Example below:

```go
func TestDiv(children ...Node) Node {
	return Div(append(
		[]Node{Class("test")},
		children...,
	)...)
}

func MainDiv() Node {
	return TestDiv(Class("main"))
}
```
Outputs invalid HTML:
```html
<div class="test" class="main"></div>
```
Should output:
```html
<div class="test main"></div>
```
I don't know if my implementation is the best way to do it, but it seems to work. Doesn't work with `Classes{}` yet though.

### DIFF
--- a/gomponents.go
+++ b/gomponents.go
@@ -79,8 +79,35 @@ func El(name string, children ...Node) Node {
 	})
 }
 
+// Prevents class="" appearing twice.
+func combineChildren(children []Node) []Node {
+	var combined []Node
+
+	var foundClassAttr *attr
+
+	for _, child := range children {
+		p, ok := child.(*attr)
+
+		if ok && p.name == "class" {
+			*p.value = strings.TrimSpace(*p.value)
+			if foundClassAttr == nil {
+				foundClassAttr = p
+			} else {
+				*foundClassAttr.value += " " + *p.value
+				continue // dont append
+			}
+		}
+
+		combined = append(combined, child)
+	}
+
+	return combined
+}
+
 func render(w2 io.Writer, name *string, children ...Node) error {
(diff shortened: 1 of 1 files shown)


markuswustenberg:
Hi @makinori, thank you for wanting to contribute!

Unfortunately, this isn't something I want to merge. I like to keep the library "dumb", so that which gets written in Go is exactly the HTML that's coming out.

If you need a higher-level component to apply classes conditionally, have a look at [`Classes`](https://pkg.go.dev/maragu.dev/gomponents@v1.1.0/components#Classes).

makinori:
I understand. It's just that the library can output invalid html

markuswustenberg:
Ah, sorry, you mentioned `Classes` yourself. 😊

> I understand. It's just that the library can output invalid html

Yes, that's actually by design. Just like you can output invalid HTML by writing invalid HTML. gomponents is not trying to validate any of the HTML coming in or out. That's a whole can of worms I don't want to open. 😅

makinori:
Ah okay, makes sense. I think it's a good idea to design and keep it that way. I think making extra helper functions for my project is probably the best option. Sorry for misunderstanding. I really do like this library haha

markuswustenberg:
No worries, I really appreciate your wanting to contribute and improve! 😊

What you would probably typically do instead is something like this:

```
func TestDiv(classes string, children ...Node) Node {
	return Div(Classes{"test":true, classes: true}, Group(children))
}

func MainDiv() Node {
	return TestDiv("main", Span())
}
```

PR #257, opened 2025-05-14
Add "search" HTML element

theandrew168:
Hey there,

Thanks for this project! Super cool idea and I've been exploring using it to replace the standard `html/template` package in one of my Go projects.

One thing I noticed, however, is that the [search element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/search) appears to be missing. This PR adds it (and updates the tests, as well).

Please let me know if I've missed anything or if any changes should be made.

Thanks again!

### DIFF
--- a/html/elements.go
+++ b/html/elements.go
@@ -256,6 +256,10 @@ func Script(children ...g.Node) g.Node {
 	return g.El("script", children...)
 }
 
+func Search(children ...g.Node) g.Node {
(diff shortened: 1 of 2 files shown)


markuswustenberg:
Hi @theandrew168, thank you for the kind words and your contribution!

I hadn't heard of the search element before. This project really keeps me up-to-date with the HTML spec. 😁

PR #262, opened 2025-06-26
Add `components.JoinAttrs`

### LINKED ISSUE #258: Investigate `class` attribute merging
I've thought about two options:

1. Have some kind of render hook that custom components can implement, and which I can check inside the `Render` function without breaking anything.
2. Wrapping the whole render process in some kind of buffered renderer that checks for duplicate `class` attributes and merges them.

markuswustenberg:
This adds `components.JoinAttrs`, a helper to join attribute values with a given name. Example:

```go
func myButton(children ...g.Node) g.Node {
	return Div(JoinAttrs("class", g.Group(children), Class("button")))
}

func myPrimaryButton(text string) g.Node {
	return myButton(Class("primary"), g.Text(text))
}

func ExampleJoinAttrs() {
	danceButton := myPrimaryButton("Dance")
	_ = danceButton.Render(os.Stdout)
	// Output: <div class="primary button">Dance</div>
}
```

Fixes #258

### DIFF
--- a/LLMs.md
+++ b/LLMs.md
@@ -1,0 +1,524 @@
+# gomponents: HTML Components in Pure Go - LLM Documentation
+
+## Overview
+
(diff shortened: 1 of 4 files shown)


markuswustenberg:
Ooh, thanks for noticing!

PR #266, opened 2025-09-02
Add MapWithIndex function and corresponding tests

MateoCaicedoW:
I use Gomponents in a couple of projects and within all of them I had to create this helper in order to get the index when iterating. So I think this is really helpful.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,10 +27,11 @@ jobs:
           - "1.22"
           - "1.23"
           - "1.24"
+          - "1.25"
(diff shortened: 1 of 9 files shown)


markuswustenberg:
Hi @MateoCaicedoW , thank you for contributing! 😊

Unfortunately, this is not something I want to add. See https://github.com/maragudk/gomponents?tab=readme-ov-file#id-like-to-add-feature-x-can-i-do-that .

Maybe you could use a collections library, like https://github.com/elliotchance/pie ?

PR #311, opened 2026-04-08
Optimize Text[f]

Hendrikto:
By making use of the new `raw` type, we can also simplify and optimize `Text[f]`, similar to the `Raw[f]` optimizations[0].

Before:
    
```
❯ go test . -bench=Text -benchmem
goos: linux
goarch: amd64
pkg: maragu.dev/gomponents
BenchmarkText/simple_text_element-32            32014112               41.18 ns/op           40 B/op          2 allocs/op
BenchmarkTextf/formatted_text_element-32        13197096               92.03 ns/op          112 B/op          4 allocs/op
PASS
ok      maragu.dev/gomponents   2.536s
```

After:

```
❯ go test . -bench=Text -benchmem
goos: linux
goarch: amd64
pkg: maragu.dev/gomponents
BenchmarkText/simple_text_element-32            32165613               39.41 ns/op           32 B/op          2 allocs/op
BenchmarkTextf/formatted_text_element-32        15473426               77.65 ns/op           64 B/op          3 allocs/op
PASS
ok      maragu.dev/gomponents   2.472s
```

[0]: https://github.com/maragudk/gomponents/pull/307

### DIFF
--- a/gomponents.go
+++ b/gomponents.go
@@ -308,26 +308,12 @@ func (a attrFunc) String() string {
 
 // Text creates a text DOM [Node] that Renders the escaped string t.
 func Text(t string) Node {
-	return NodeFunc(func(w io.Writer) error {
-		if w, ok := w.(io.StringWriter); ok {
-			_, err := w.WriteString(template.HTMLEscapeString(t))
-			return err
-		}
-		_, err := w.Write([]byte(template.HTMLEscapeString(t)))
-		return err
-	})
+	return raw(template.HTMLEscapeString(t))
 }
 
 // Textf creates a text DOM [Node] that Renders the interpolated and escaped string format.
 func Textf(format string, a ...interface{}) Node {
-	return NodeFunc(func(w io.Writer) error {
-		if w, ok := w.(io.StringWriter); ok {
-			_, err := w.WriteString(template.HTMLEscapeString(fmt.Sprintf(format, a...)))
-			return err
-		}
-		_, err := w.Write([]byte(template.HTMLEscapeString(fmt.Sprintf(format, a...))))
-		return err
-	})
+	return raw(template.HTMLEscapeString(fmt.Sprintf(format, a...)))
 }
 
 // Compile-time check that [raw] implements [fmt.Stringer] and [Node].
--- a/gomponents.go
+++ b/gomponents.go
@@ -316,10 +316,11 @@ func Textf(format string, a ...interface{}) Node {
 	return raw(template.HTMLEscapeString(fmt.Sprintf(format, a...)))
 }
 
-// Compile-time check that [raw] implements [fmt.Stringer] and [Node].
+// Compile-time check that [raw] implements [fmt.Stringer], [Node], and [nodeTypeDescriber].
 var _ interface {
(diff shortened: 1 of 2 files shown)


gomponents.go
```diff
@@ -308,26 +308,12 @@ func (a attrFunc) String() string {
 
 // Text creates a text DOM [Node] that Renders the escaped string t.
 func Text(t string) Node {
-	return NodeFunc(func(w io.Writer) error {
```
markuswustenberg, line 311: Something I didn't think about before is that this is actually also a breaking change, since `NodeFunc` is public, and type assertions like `Text("").(NodeFunc)` will now fail. I really don't see why anyone would do that though, so I think that's an acceptable risk. We're on the edge of compatibility. :-) So just a comment, nothing to do about this.

gomponents.go
markuswustenberg, line 322: Maybe it should also satisfy `nodeTypeDescriber` like `NodeFunc` did? Just for completeness, although it's functionally the same.

markuswustenberg (approved):
Excellent, thank you again!

gomponents.go
Hendrikto: This is something I did consider, and I think this is fine, because all of the return types have always been just `g.Node`. Relying on them being anything other than that, like with an unconditional type assertion, was always relying on internal details that were never promised by the API.

It is conceivable somebody might have done something like that, although I do not see a reason why, but that was always bad practice and fragile.

gomponents.go
Hendrikto: The explicitly documented default `NodeType` is `ElementType`, so it is not necessary, as you said yourself. However, I also do not see big drawbacks. Implementing `nodeTypeDescriber` is a bit more code that does not really have any effect, but it is also a little more explicit than relying on the default :shrug: I am not strongly leaning either way, I think both are roughly equivalent.

Were you just thinking out loud, or should I take action?

gomponents.go
markuswustenberg: Yeah, agreed.

gomponents.go
markuswustenberg: I think we should implement it for completeness here. That's another _technically_ a breaking change that probably no-one relies on, but if we can avoid those easily, I think we should. :-)

Hendrikto: done

markuswustenberg:
Nice, thank you again for your work!

Anything else in your pipeline? 😁

Hendrikto:
> Anything else in your pipeline? 😁

Yes, I do have a few more ideas. I think it is possible to optimize more things, although I will have to get some client work done first 😅

I will get back to you in a few days.

markuswustenberg:
What?! I fully expected you to work on this full time now. 😁

Take all the time you need of course. Looking forward to seeing what you've got.

Hendrikto:
> What?! I fully expected you to work on this full time now. 😁

I am open to that possibility. I just need to figure something out regarding paying my bills first :laughing:

markuswustenberg:
> I am open to that possibility. I just need to figure something out regarding paying my bills first 😆

Oh yeah, money. I've heard about that.

PR #321, opened 2026-04-14
Add missing track element

Hendrikto:
I noticed that the [`track` element](https://www.w3.org/TR/2011/WD-html-markup-20110113/track.html) is missing. It is already listed in `isVoidElement`, but not actually implemented itself.

### DIFF
--- a/html/elements.go
+++ b/html/elements.go
@@ -469,3 +469,7 @@ func Var(children ...g.Node) g.Node {
 func Video(children ...g.Node) g.Node {
 	return g.El("video", g.Group(children))
 }
+
+func Track(children ...g.Node) g.Node {
+	return g.El("track", children...)
+}
--- a/html/elements_test.go
+++ b/html/elements_test.go
@@ -157,6 +157,7 @@ func TestSimpleVoidKindElements(t *testing.T) {
 		{Name: "param", Func: Param},
 		{Name: "source", Func: Source},
 		{Name: "wbr", Func: Wbr},
+		{Name: "track", Func: Track},
 	}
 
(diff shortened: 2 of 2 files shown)


Hendrikto:
I also noticed that from the other elements listed in `isVoidElement`, [`command`](https://www.w3.org/TR/2011/WD-html-markup-20110113/command.html) and [`keygen`](https://www.w3.org/TR/2011/WD-html-markup-20110113/keygen.html) are not implemented either. I do not think they should be, as both are deprecated[0] [1], but we could remove them from the `isVoidElement` check to improve efficiency.

[0]: https://udn.realityripple.com/docs/Web/HTML/Element/keygen
[1]: https://udn.realityripple.com/docs/Web/HTML/Element/command

markuswustenberg:
> I also noticed that from the other elements listed in `isVoidElement`, [`command`](https://www.w3.org/TR/2011/WD-html-markup-20110113/command.html) and [`keygen`](https://www.w3.org/TR/2011/WD-html-markup-20110113/keygen.html) are not implemented either. I do not think they should be, as both are deprecated[0](https://udn.realityripple.com/docs/Web/HTML/Element/keygen) [1](https://udn.realityripple.com/docs/Web/HTML/Element/command), but we could remove them from the `isVoidElement` check to improve efficiency.

I think it makes sense to keep them there because people can still use the `El` function to create elements, and they should still be handled correctly as void elements, even though they're deprecated.

Hendrikto:
> I think it makes sense to keep them there because people can still use the `El` function to create elements

You are right, I did not fully think it through.

PR #322, opened 2026-04-20
Use io.WriteString

Hendrikto:
The code can be simplified by using `io.WriteString`.

### DIFF
--- a/gomponents.go
+++ b/gomponents.go
@@ -88,31 +88,21 @@ var (
 // Use this if no convenience creator exists in the html package.
 func El(name string, children ...Node) Node {
 	return NodeFunc(func(w io.Writer) error {
-		var err error
-
-		sw, ok := w.(io.StringWriter)
-
-		if _, err = w.Write(lt); err != nil {
+		if _, err := w.Write(lt); err != nil {
 			return err
 		}
 
-		if ok {
-			if _, err = sw.WriteString(name); err != nil {
-				return err
-			}
-		} else {
-			if _, err = w.Write([]byte(name)); err != nil {
-				return err
-			}
+		if _, err := io.WriteString(w, name); err != nil {
+			return err
 		}
 
 		for _, c := range children {
-			if err = renderChild(w, c, AttributeType); err != nil {
+			if err := renderChild(w, c, AttributeType); err != nil {
 				return err
 			}
 		}
 
-		if _, err = w.Write(gt); err != nil {
+		if _, err := w.Write(gt); err != nil {
 			return err
 		}
 
@@ -121,26 +111,20 @@ func El(name string, children ...Node) Node {
 		}
 
 		for _, c := range children {
-			if err = renderChild(w, c, ElementType); err != nil {
+			if err := renderChild(w, c, ElementType); err != nil {
 				return err
 			}
 		}
 
-		if _, err = w.Write(ltSlash); err != nil {
+		if _, err := w.Write(ltSlash); err != nil {
 			return err
 		}
 
-		if ok {
-			if _, err = sw.WriteString(name); err != nil {
-				return err
-			}
-		} else {
-			if _, err = w.Write([]byte(name)); err != nil {
-				return err
-			}
+		if _, err := io.WriteString(w, name); err != nil {
+			return err
 		}
 
-		if _, err = w.Write(gt); err != nil {
+		if _, err := w.Write(gt); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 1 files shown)


Hendrikto:
I am not sure about [`El`](https://github.com/Hendrikto/gomponents/blob/io.WriteString/gomponents.go#L89) and [`valueAttr`](https://github.com/Hendrikto/gomponents/blob/io.WriteString/gomponents.go#L233). We could also use `io.StringWriter` there, simplifying the code, but then the type assertion `w.(io.StringWriter)` would happen twice. This probably does not make a performance difference in practice, so it might be worth it for the easier to read code. Let me know what you think.

markuswustenberg:
I actually don't remember why I did it this way, but note that with this change, the interface satisfaction check happens on each write. Not sure if that has any implications in practice. Have you measured?

Hendrikto:
> the interface satisfaction check happens on each write

If I understand you correctly, this is exactly what I referred to here:

> We could also use io.StringWriter [for [El](https://github.com/Hendrikto/gomponents/blob/io.WriteString/gomponents.go#L89) and [valueAttr](https://github.com/Hendrikto/gomponents/blob/io.WriteString/gomponents.go#L233)], […] but then the type assertion `w.(io.StringWriter)` would happen twice.

That is why I limited the changes to places where there is only a single write, and did not touch `El` and `valueAttr` for now, so the semantics are fully compatible.

> Not sure if that has any implications in practice. Have you measured?

I measured the current commit with only single write cases, and there is no difference at all. This is expected, as the semantics are exactly the same.

I did not yet measure the multiple write cases that are not currently included in this PR. If you agree with the changes in principle, I can measure that too.

markuswustenberg:
> > the interface satisfaction check happens on each write
> 
> If I understand you correctly, this is exactly what I referred to here:
> 
> > We could also use io.StringWriter [for [El](https://github.com/Hendrikto/gomponents/blob/io.WriteString/gomponents.go#L89) and [valueAttr](https://github.com/Hendrikto/gomponents/blob/io.WriteString/gomponents.go#L233)], […] but then the type assertion `w.(io.StringWriter)` would happen twice.
> 
> That is why I limited the changes to places where there is only a single write, and did not touch `El` and `valueAttr` for now, so the semantics are fully compatible.

Ah, right, of course. Honestly, I cannot imagine that it would have any performance impact in practice. Compared to I/O and memory stuff, checking interfaces must be pretty far down that list.

> > Not sure if that has any implications in practice. Have you measured?
> 
> I measured the current commit with only single write cases, and there is no difference at all. This is expected, as the semantics are exactly the same.
> 
> I did not yet measure the multiple write cases that are not currently included in this PR. If you agree with the changes in principle, I can measure that too.

Let's just include them in this PR, makes the code simpler.

Hendrikto:
> Let's just include them in this PR, makes the code simpler.

Okay, I did include the multi-write cases as well. As we expected, it does not make a measurable performance difference either way. The results are all within margin of error:

Before:
    
```
goos: linux
goarch: amd64
pkg: maragu.dev/gomponents
BenchmarkAttr/boolean_attributes-32             44900137               28.63 ns/op           32 B/op          2 allocs/op
BenchmarkAttr/name-value_attributes-32          17749360               68.05 ns/op           72 B/op          3 allocs/op
BenchmarkEl/normal_elements-32                  19974408               59.38 ns/op           72 B/op          3 allocs/op
BenchmarkRaw/raw_element-32                     64644760               17.17 ns/op           24 B/op          1 allocs/op
BenchmarkRawf/formatted_raw_element-32          18530452               61.89 ns/op           64 B/op          3 allocs/op
BenchmarkText/simple_text_element-32            29606492               39.96 ns/op           32 B/op          2 allocs/op
BenchmarkTextf/formatted_text_element-32        15062982               79.14 ns/op           64 B/op          3 allocs/op

BenchmarkRealisticPage/construct_and_render-32             7206    145050 ns/op  342092 B/op       3068 allocs/op
BenchmarkRealisticPage/render_pre-built_tree-32           13208     90683 ns/op  213120 B/op         39 allocs/op
```

After

```
goos: linux
goarch: amd64
pkg: maragu.dev/gomponents
BenchmarkAttr/boolean_attributes-32             46899280               28.31 ns/op           32 B/op          2 allocs/op
BenchmarkAttr/name-value_attributes-32          16446249               69.27 ns/op           72 B/op          3 allocs/op
BenchmarkEl/normal_elements-32                  19708002               60.78 ns/op           72 B/op          3 allocs/op
BenchmarkRaw/raw_element-32                     67244452               17.33 ns/op           24 B/op          1 allocs/op
BenchmarkRawf/formatted_raw_element-32          19076622               62.39 ns/op           64 B/op          3 allocs/op
BenchmarkText/simple_text_element-32            31093843               39.89 ns/op           32 B/op          2 allocs/op
BenchmarkTextf/formatted_text_element-32        15574497               79.21 ns/op           64 B/op          3 allocs/op

BenchmarkRealisticPage/construct_and_render-32             8074    154903 ns/op  342085 B/op       3068 allocs/op
BenchmarkRealisticPage/render_pre-built_tree-32           13353     89538 ns/op  213120 B/op         39 allocs/op
```

markuswustenberg:
Excellent, thank you. Okay to merge?

Hendrikto:
> Okay to merge?

I think so. All tests are passing.

markuswustenberg:
Thanks again! 😊

Issue #323, opened 2026-05-30
feature: Static function to render components once


guillemus:
So, because of how gomponents work, there's a lot of work to be done for large static components with lots of strings and attributes.

This work is redundant if there's no dynamic data in between those nodes when rendering multiple times. And where this matters, there's a lot to gain in terms of performance, you get very close to templ / other libraries.

It's similar to "memoizing" the component, or caching the rendering, or the static render props thingy that nextjs does (I believe?).

The API is dead simple:

```go
package views

import (
	"io"
	"strings"
	"sync"

	. "maragu.dev/gomponents"
)

func Static(node Node) Node {
	var once sync.Once
	var html string
	var renderErr error

	return NodeFunc(func(w io.Writer) error {
		once.Do(func() {
			var b strings.Builder
			renderErr = node.Render(&b)
			html = b.String()
		})
		if renderErr != nil {
			return renderErr
		}
		_, writeErr := io.WriteString(w, html)
		return writeErr
	})
}

```

And an example of the usage.

```go
var headStatic = Static(Group{
	Link(Rel("preload"), Href("/public/fonts/figtree-latin-var.woff2"), As("font"), Type("font/woff2"), Attr("crossorigin")),
	Link(Rel("preload"), Href("/public/fonts/jetbrains-mono-latin-var.woff2"), As("font"), Type("font/woff2"), Attr("crossorigin")),
	Link(Rel("stylesheet"), Href("https://cdn.jsdelivr.net/npm/katex@0.16.11/dist/katex.min.css")),
	Script(Type("module"), Src("https://cdn.jsdelivr.net/gh/starfederation/datastar@v1.0.1/bundles/datastar.js")),
	Script(Type("speculationrules"), Raw(`{
		"prefetch": [{
			"where": { "href_matches": "/*" },
			"eagerness": "moderate"
		}]
	}`)),
})
```

What do you think @markuswustenberg? I'm open to create the PR + tests.

markuswustenberg:
I've thought of doing this myself a couple of times, but I always end up not bothering, because I'm not convinced it's worth it performance-wise. In all my years of using gomponents, I've never had a (measurable) need. And even if that were the case, the caching layer is often elsewhere than in the Go code.

_That said_, I'm open to giving it a try, because I of course know that everyone's needs are not equal.

I'm curious, was your issue born out of a specific need? And if, was it for purely static content, or depending on _some_ state (think the `Vary` HTTP header)?

guillemus:
[...]

PR #325, opened 2026-06-05
Add "Static" func

guillemus:
As discussed here #323 , here's the Static implementation.

@markuswustenberg I did add the static func at "gomponents.go". It feels enough of a primitive to be there I think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,8 @@ Check out [www.gomponents.com](https://www.gomponents.com) for an introduction o
   - `Text` and `Textf` that insert HTML-escaped text,
   - `Raw` and `Rawf` for inserting raw strings,
   - `Map` for mapping data to components and `Group` for grouping components,
-  - and `If`/`Iff` for conditional rendering.
+  - `If`/`Iff` for conditional rendering,
+  - and `Static` for caching large static element trees.
 - No external dependencies
 - Mature and stable, no breaking changes
 
@@ -93,6 +94,7 @@ When you create a void element, any child nodes that are not attributes will be 
 
 gomponents renders directly to an `io.Writer`, making it efficient for server-side rendering.
 The library avoids unnecessary allocations where possible.
+For large element trees that never depend on request data, user data, or changing state, wrap them in `Static` once at package scope and reuse the node. `Static` is useful only when the returned node is reused.
 
 ## FAQ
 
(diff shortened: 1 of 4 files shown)


gomponents_test.go
```diff
@@ -387,3 +390,89 @@ func ExampleIff() {
...
+			t.Fatalf("expected 1 render, got %v", atomic.LoadInt64(&renders))
+		}
+	})
+}
+
+func ExampleStatic() {
+	staticHead := g.Static(h.Head(
+		h.TitleEl(g.Text("My site")),
+		h.Link(h.Rel("stylesheet"), h.Href("/app.css")),
+	))
```
markuswustenberg, line 473: I've been thinking a bit about the usage of this. It's a shame you lose the ability to inline the `Static` component like you usually would with all the other components, and that you have to keep the node variable (`staticHead` here) somewhere as state between requests.

Would it be nicer to perhaps pass a string pointer or something, that the receiving function renders to if non-empty? And subsequent requests see the non-empty pointer value and just use that.

README.md
```diff
@@ -36,7 +36,8 @@ Check out [www.gomponents.com](https://www.gomponents.com) for an introduction o
   - `Text` and `Textf` that insert HTML-escaped text,
   - `Raw` and `Rawf` for inserting raw strings,
   - `Map` for mapping data to components and `Group` for grouping components,
-  - and `If`/`Iff` for conditional rendering.
+  - `If`/`Iff` for conditional rendering,
+  - and `Static` for caching large static element trees.
```
markuswustenberg, line 40: I definitely rather want this in the `components` package, like the other specialized helpers there. I don't see it as a low-level primitive. Could you move it there?

markuswustenberg (commented):
Let's discuss the implementation details a bit more, if you're up for it. :-)

If you'd rather not, feel free to let me know, and we can merge and I'll iterate myself.

guillemus:
Hey @markuswustenberg ! there will probably be lots of back and forth, and this pr requires more taste than tedious work really. The code by itself is very small.

Feel free to merge this / close this pr and implement it the way you want it. I thought more about the global key val cache that you proposed, could be another interesting approach, I might have overrated the complexity.

markuswustenberg:
Okay, thank you for the reply and your work so far. I'll try some stuff out when I have some spare time. :-)

PR #329, opened 2026-07-28
Add ruby, rt, and rp elements

milanobrtlik:
The ruby, rt, and rp elements are missing from the html package.  

They go together: rp provides the parenthesis fallback for browsers that don't render rt annotations, so adding one without the others isn't much use.  

Spec: https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-ruby-element


markuswustenberg:
I did not know about those! Thank you. 😊

PR #335, opened 2026-07-29
Fix `JoinAttrs` not joining attributes in nested groups

milanobrtlik:
`JoinAttrs` unwraps `g.Group` only one level deep, so an attribute nested deeper is never
merged. It falls through to the result verbatim, and `renderChild`, which recurses into groups
at any depth, then emits it as a second attribute of the same name. Browsers keep the first
occurrence, so the value is silently dropped.

### Reproduction

Against `main` @ d1fd82b:

```go
Div(JoinAttrs("class", g.Group{Class("party"), g.Group{Class("gold")}}, Class("hat")))
```

renders

```html
<div class="party hat" class="gold"></div>
```

instead of `<div class="party gold hat"></div>`. With three levels it degrades further —
`<div class="a" class="b" class="c d"></div>`.

### Why it comes up

Two-level groups arise from the composition pattern the docs already show: a component that
forwards `g.Group(children)` into `JoinAttrs`, called by another component that does the same.
`g.Map` also returns a `Group`, so mapping over data into a `JoinAttrs` call hits it too. The
existing group test missed this because it forwards an element rather than an attribute, and
elements are appended to `result` either way.

### Fix

Unwrap recursively in `processNode`, mirroring what `renderChild` does, so `JoinAttrs` sees
exactly what will be rendered. Element order is preserved: non-attribute children stay in place
and the joined attribute lands at the position of the first match, same as before.

Four test cases added — two-level, three-level, nested components, and element order inside
nested groups. Existing tests unchanged, coverage stays at 100%.

The doc comment is updated to say groups are transparent, since the previous wording
("Attributes on non-direct descendants are ignored") no longer describes the behaviour.

### DIFF
--- a/components/components.go
+++ b/components/components.go
@@ -67,14 +67,17 @@ func (c Classes) String() string {
 }
 
 // JoinAttrs joins attributes with the given name on the first level of the given nodes.
-// Attributes on non-direct descendants are ignored.
+// A [g.Group] is transparent and doesn't count as a level, so attributes inside one are joined
+// at any depth, matching how groups are rendered.
+// Attributes on non-direct descendants, for example inside a child element, are ignored.
 // Non-empty attribute values are joined by spaces into a single attribute.
 // Empty, whitespace-only, and boolean (valueless) attributes are deduplicated and discarded
 // if a non-empty value exists. If only boolean/empty attributes match, a single boolean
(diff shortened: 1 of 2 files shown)


markuswustenberg:
Thanks! I had forgotten about this one. :-)

PR #345, opened 2026-09-01
Add the missing global attributes

milanobrtlik:
`html/attributes.go` was missing 11 of the global attributes in the [WHATWG attribute index](https://html.spec.whatwg.org/multipage/indices.html#attributes-3). Since they apply to every element, they come up a lot in real use.

Added: `accesskey`, `autocapitalize`, `autocorrect`, `contenteditable`, `enterkeyhint`, `inert`, `inputmode`, `is`, `nonce`, `translate`, `writingsuggestions`.

A few notes:

- `inert` is the only boolean one, so it's the only one without a value parameter. I checked each against the "Value" column of the spec's attribute index rather than going by feel, since that's the one place this kind of change can actually be wrong.
- None of the names clash with an existing element or attribute, so none need an `Attr` suffix and the list in the README is unchanged.
- Inserted alphabetically into both blocks in `attributes.go` and both tables in `attributes_test.go`.
- `html` stays at 100% coverage, and `golangci-lint run` is clean.

I left out `headingoffset` and `headingreset` on purpose. They're conforming in the index, but they're much newer than the rest and I'd rather not have them hold up the other 11. Happy to send them as a separate PR if you want them.


markuswustenberg:
Excellent, thanks!

PR #346, opened 2026-09-01
Benchmark against writers that don't allocate per render

milanobrtlik:
The benchmarks render into a `strings.Builder` and call `Reset` between iterations. `strings.Builder.Reset` drops the underlying buffer rather than keeping it, so every iteration re-grows it from nil, and the reported numbers include the allocator work of the benchmark harness itself rather than just the library's.

`bytes.Buffer.Reset` keeps its capacity, which isolates the code under test. It's also closer to real usage, where output goes to a writer that isn't re-allocated per render.

`BenchmarkRealisticPage`, `-count=5`, go1.24.10, linux/amd64:

| | B/op | ns/op |
|---|---|---|
| construct and render | 342216 → 130325 (-62%) | 182366 → 145494 (-20%) |
| render pre-built tree | 213556 → 1863 (-99%) | 114446 → 66015 (-42%) |

The micro benchmarks are affected the same way — `BenchmarkAttr/name-value` and `BenchmarkEl/normal_elements` drop from 3 allocs/op to 1, and `BenchmarkAttr/boolean` and `BenchmarkText/simple` from 2 to 1. The library only ever made one; the rest was the builder re-growing.

For the pre-built tree case that means 99% of the memory being measured was the harness, not the render.

Worth noting that this makes the numbers incomparable to any recorded before it, so it's a one-off break in the series. Test-only change, no library code touched.

---

**Update:** after review by @Hendrikto and @markuswustenberg, the approach changed. The per-attribute and per-element benchmarks now render to `io.Discard`, and `BenchmarkRealisticPage` keeps its construct/pre-built split while gaining a `discarded`/`buffered` writer axis, where `buffered` is a `bufio.Writer` of 2048 bytes — the size `net/http` puts in front of a handler's `ResponseWriter`. Numbers and reasoning in [this comment](https://github.com/maragudk/gomponents/pull/346#issuecomment-5526172524). Everything above is the original proposal, kept as it was so the discussion still reads.

### DIFF
--- a/gomponents_benchmark_test.go
+++ b/gomponents_benchmark_test.go
@@ -3,7 +3,7 @@
 package gomponents_test
 
 import (
-	"strings"
+	"io"
 	"testing"
 
 	g "maragu.dev/gomponents"
@@ -11,82 +11,61 @@ import (
 
 func BenchmarkAttr(b *testing.B) {
 	b.Run("boolean attributes", func(b *testing.B) {
-		var sb strings.Builder
-
 		for b.Loop() {
 			a := g.Attr("hat")
-			_ = a.Render(&sb)
-			sb.Reset()
+			_ = a.Render(io.Discard)
 		}
 	})
 
 	b.Run("name-value attributes", func(b *testing.B) {
-		var sb strings.Builder
-
 		for b.Loop() {
 			a := g.Attr("hat", "party")
-			_ = a.Render(&sb)
-			sb.Reset()
+			_ = a.Render(io.Discard)
 		}
 	})
 }
 
 func BenchmarkEl(b *testing.B) {
 	b.Run("normal elements", func(b *testing.B) {
-		var sb strings.Builder
-
 		for b.Loop() {
 			e := g.El("div")
-			_ = e.Render(&sb)
-			sb.Reset()
+			_ = e.Render(io.Discard)
 		}
 	})
 }
 
 func BenchmarkRaw(b *testing.B) {
 	b.Run("raw element", func(b *testing.B) {
-		var sb strings.Builder
-
 		for b.Loop() {
 			e := g.Raw("<span>content</span>")
-			_ = e.Render(&sb)
-			sb.Reset()
+			_ = e.Render(io.Discard)
 		}
 	})
 }
 
 func BenchmarkRawf(b *testing.B) {
 	b.Run("formatted raw element", func(b *testing.B) {
-		var sb strings.Builder
-
 		for b.Loop() {
 			e := g.Rawf("<span>%s</span>", "content")
-			_ = e.Render(&sb)
-			sb.Reset()
+			_ = e.Render(io.Discard)
 		}
 	})
 }
 
 func BenchmarkText(b *testing.B) {
 	b.Run("simple text element", func(b *testing.B) {
-		var sb strings.Builder
-
 		for b.Loop() {
 			e := g.Text("some simple text")
-			_ = e.Render(&sb)
-			sb.Reset()
+			_ = e.Render(io.Discard)
 		}
 	})
 }
 
 func BenchmarkTextf(b *testing.B) {
 	b.Run("formatted text element", func(b *testing.B) {
-		var sb strings.Builder
-
 		for b.Loop() {
 			e := g.Textf("some %s text", "formatted")
-			_ = e.Render(&sb)
-			sb.Reset()
+			_ = e.Render(io.Discard)
 		}
 	})
 }
--- a/html/benchmark_test.go
+++ b/html/benchmark_test.go
@@ -3,14 +3,19 @@
 package html_test
 
 import (
+	"bufio"
 	"fmt"
-	"strings"
+	"io"
 	"testing"
 
 	g "maragu.dev/gomponents"
 	c "maragu.dev/gomponents/components"
 	. "maragu.dev/gomponents/html"
 )
+
+// responseBufferSize is the size of the [bufio.Writer] that net/http puts in front of a
+// handler's [http.ResponseWriter], its bufferBeforeChunkingSize.
+const responseBufferSize = 2048
 
 // BenchmarkRealisticPage benchmarks rendering a full, realistic HTML page
 // resembling a typical web application dashboard with navigation, sidebar,
@@ -275,19 +280,36 @@ func BenchmarkRealisticPage(b *testing.B) {
 	}
 
 	b.Run("construct and render", func(b *testing.B) {
-		var sb strings.Builder
-		for b.Loop() {
-			_ = page().Render(&sb)
-			sb.Reset()
-		}
+		b.Run("discarded", func(b *testing.B) {
+			for b.Loop() {
+				_ = page().Render(io.Discard)
+			}
+		})
+
+		b.Run("buffered", func(b *testing.B) {
+			w := bufio.NewWriterSize(io.Discard, responseBufferSize)
+			for b.Loop() {
+				_ = page().Render(w)
+				_ = w.Flush()
+			}
+		})
 	})
 
 	b.Run("render pre-built tree", func(b *testing.B) {
-		var sb strings.Builder
 		p := page()
-		for b.Loop() {
-			_ = p.Render(&sb)
-			sb.Reset()
-		}
+
+		b.Run("discarded", func(b *testing.B) {
+			for b.Loop() {
+				_ = p.Render(io.Discard)
+			}
+		})
+
+		b.Run("buffered", func(b *testing.B) {
+			w := bufio.NewWriterSize(io.Discard, responseBufferSize)
+			for b.Loop() {
+				_ = p.Render(w)
+				_ = w.Flush()
+			}
+		})
 	})
 }


markuswustenberg:
@Hendrikto you were involved in all the benchmark stuff a while ago. If you're interested and have time, what do you think of this change?

Hendrikto:
Thank you @milanobrtlik for the high-quality PR, and thanks to @markuswustenberg for the ping. I do indeed have some input on this.

After taking a closer look, I generally agree with @milanobrtlik that the current benchmarks measure a lot of peripheral harness stuff, instead of being isolated to gomponents itself. Whether that is desirable is another question, but I tend to agree with @milanobrtlik that we should probably isolate the benchmarks to gomponents itself. I will also explain that position and expand on this in a future comment.

I do, however, think that this PR is not the most idiomatic and cleanest solution to that problem. I do have another suggestion that addresses this problem more directly and completely, while also being more idiomatic.

I do not have a lot of time right now. I will get back to you with a more detailed write-up and my suggestion probably this evening. If I do not manage, I will get back to this at the latest tomorrow.

Best regards,
Hendrik

milanobrtlik:
Thanks for taking the time, @Hendrikto, and no rush at all — I'd much rather wait for the right fix than land a mediocre one quickly.

Good to hear the diagnosis holds up independently. That was the part I was least sure of, since I only had my own profiles to go on, and it's easy to talk yourself into a number when you're the one who picked the benchmark.

I'm genuinely looking forward to your suggestion. If there's a cleaner and more complete way to isolate this, I'd rather have that than what's in this PR, so I'll leave it untouched until your write-up lands. The one thing I went back and forth on, in case it's useful context: how much a benchmark should still resemble real usage, where output does go somewhere, against isolating gomponents as completely as possible. `bytes.Buffer` was my attempt at a middle ground, but I can see the argument that isolation should simply win in a benchmark, and I don't hold that view strongly.

markuswustenberg:
Hey @Hendrikto, no rush at all! I’m just happy you’re willing to take a look. :-) Take as long as you need!

Hendrikto:
Thanks for the kind words and patience @markuswustenberg and @milanobrtlik. Here are my expanded thoughts, explanations, and suggestions:

## Reproducing the Results

First of all, I can reproduce @milanobrtlik's results:

<details>
<summary>maragudk/main</summary>

```
❯ git rev-parse HEAD
ff049a48e979902695e6096be541fc36197ba949 # maragudk/main
goos: linux
goarch: amd64
pkg: maragu.dev/gomponents
BenchmarkAttr/boolean_attributes-32         	55663402	       22.57 ns/op	     32 B/op	      2 allocs/op
BenchmarkAttr/name-value_attributes-32      	19224150	       62.95 ns/op	     72 B/op	      3 allocs/op
BenchmarkEl/normal_elements-32              	23068840	       52.14 ns/op	     72 B/op	      3 allocs/op
BenchmarkRaw/raw_element-32                 	81590991	       14.97 ns/op	     24 B/op	      1 allocs/op
BenchmarkRawf/formatted_raw_element-32      	22077214	       53.50 ns/op	     64 B/op	      3 allocs/op
BenchmarkText/simple_text_element-32        	41122095	       29.49 ns/op	     32 B/op	      2 allocs/op
BenchmarkTextf/formatted_text_element-32    	19653206	       62.59 ns/op	     64 B/op	      3 allocs/op
PASS
ok  	maragu.dev/gomponents	8.520s
PASS
ok  	maragu.dev/gomponents/components	0.002s
goos: linux
goarch: amd64
BenchmarkRealisticPage/construct_and_render-32         	   9156	   113158 ns/op	 342088 B/op	   3068 allocs/op
BenchmarkRealisticPage/render_pre-built_tree-32        	  16226	    73901 ns/op	 213120 B/op	     39 allocs/op
PASS
ok  	maragu.dev/gomponents/html	2.239s
PASS
ok  	maragu.dev/gomponents/http	0.002s
?   	maragu.dev/gomponents/internal/assert	[no test files]
PASS
ok  	maragu.dev/gomponents/internal/import	0.002s
PASS
ok  	maragu.dev/gomponents/x/slices	0.001s
go test ./... -bench=. -benchmem  13.10s user 0.72s system 126% cpu 10.899 total
```
</details>

<details>
<summary>milanobrtlik/benchmark-buffer</summary>

```
❯ git rev-parse HEAD
e08c49c30cfb0bcbe45a8740d1302f3ba5fd3755 # milanobrtlik/benchmark-buffer
goos: linux
goarch: amd64
pkg: maragu.dev/gomponents
BenchmarkAttr/boolean_attributes-32         	97109070	       13.30 ns/op	     24 B/op	      1 allocs/op
BenchmarkAttr/name-value_attributes-32      	33060132	       38.11 ns/op	     48 B/op	      1 allocs/op
BenchmarkEl/normal_elements-32              	47389102	       24.90 ns/op	     48 B/op	      1 allocs/op
BenchmarkRaw/raw_element-32                 	294751494	        4.069 ns/op	      0 B/op	      0 allocs/op
BenchmarkRawf/formatted_raw_element-32      	27966744	       42.91 ns/op	     40 B/op	      2 allocs/op
BenchmarkText/simple_text_element-32        	60379384	       20.11 ns/op	     16 B/op	      1 allocs/op
BenchmarkTextf/formatted_text_element-32    	23962275	       50.31 ns/op	     40 B/op	      2 allocs/op
PASS
ok  	maragu.dev/gomponents	8.557s
PASS
ok  	maragu.dev/gomponents/components	0.002s
goos: linux
goarch: amd64
BenchmarkRealisticPage/construct_and_render-32         	  10000	   102406 ns/op	 129983 B/op	   3048 allocs/op
BenchmarkRealisticPage/render_pre-built_tree-32        	  20464	    58318 ns/op	   1430 B/op	     19 allocs/op
PASS
ok  	maragu.dev/gomponents/html	2.221s
PASS
ok  	maragu.dev/gomponents/http	0.002s
?   	maragu.dev/gomponents/internal/assert	[no test files]
PASS
ok  	maragu.dev/gomponents/internal/import	0.002s
PASS
ok  	maragu.dev/gomponents/x/slices	0.002s
go test ./... -bench=. -benchmem  12.98s user 0.53s system 121% cpu 11.111 total
```
</details>

<details>
<summary>comparison</summary>

| Benchmark                                    | maragudk/main               | milanobrtlik/benchmark-buffer |
|----------------------------------------------|-----------------------------|-------------------------------|
| BenchmarkAttr/boolean_attributes             |     32 B/op,    2 allocs/op |     24 B/op,    1 allocs/op   |
| BenchmarkAttr/name                           |     72 B/op,    3 allocs/op |     48 B/op,    1 allocs/op   |
| BenchmarkEl/normal_elements                  |     72 B/op,    3 allocs/op |     48 B/op,    1 allocs/op   |
| BenchmarkRaw/raw_element                     |     24 B/op,    1 allocs/op |      0 B/op,    0 allocs/op   |
| BenchmarkRawf/formatted_raw_element          |     64 B/op,    3 allocs/op |     40 B/op,    2 allocs/op   |
| BenchmarkText/simple_text_element            |     32 B/op,    2 allocs/op |     16 B/op,    1 allocs/op   |
| BenchmarkTextf/formatted_text_element        |     64 B/op,    3 allocs/op |     40 B/op,    2 allocs/op   |
| BenchmarkRealisticPage/construct_and_render  | 342088 B/op, 3068 allocs/op | 129983 B/op, 3048 allocs/op   |
| BenchmarkRealisticPage/render_pre-built_tree | 213120 B/op,   39 allocs/op |   1430 B/op,   19 allocs/op   |
</details>

The numbers do not match exactly, which probably has to do with the [recent allocation optimizations](https://go.dev/blog/allocation-optimizations), and maybe [also this](https://go.dev/doc/go1.27#faster-memory-allocation) (I am on `go1.27.0-X:nodwarf5`), but they are very close. Also, for the larger tests in `maragu.dev/gomponents/html`, I have always observed a slight variability between runs.

## What is the Intent?

The main question, in my opinion, then becomes what the intent of the benchmarks is. I tend to agree with @milanobrtlik (if I understand him correctly), that the benchmarks should focus on the part we have direct control over, in order to focus on the library code itself, and get rid of the noise surrounding that. The new benchmarks results, for example, correctly show that after that `g.Raw` optimizations, a raw element can be rendered without any memory allocations, which was previously shadowed by the internal `strings.Builder` allocation.

Another example, where the new benchmarks show what is going on much clearer, are the comparison between `BenchmarkRealisticPage/construct_and_render` and `BenchmarkRealisticPage/render_pre-built_tree`. The advantage of pre-rendering is actually huge, but the old benchmarks did not clearly show that, as a significant part of the time and allocations came from just the harness.

If you agree with the assumption that the benchmarks should focus on the library code itself, and not contain noise from outside of our control (i.e. what happens within the `io.Writer` we render to), I do have a suggestion for accomplishing that more completely and idiomatically. More on that below.

One could also argue that we should consider real world interactions with the `io.Writer` we render to, since that is the reason for using `io.WriteString` for instance. That is a performance optimization targeted at the `io.Writer`, otherwise we could just always cast strings to `[]byte` and `Write()` that directly, which would simplify the code and accomplish the same result. The sole reason for using the `io.StringWriter` interface, when available, is external optimization.

However, I would argue that we never had a reference value, since we always rendered to `strings.Builder`, which satisfies `io.StringWriter`, so we always, unconditionally used the `WriteString()` method. Thinking about this, it is unclear whether that optimization even makes sense in the first place. We have nothing to compare to. Moreover, rendering to a `strings.Builder` is probably not very "real world" anyway, especially when `Reset()`ting if after every write.

## Real World?

One thing I had to think about for a bit is the claim that

> It's also closer to real usage

I definitely agree that rendering a single element or attribute and then immediately discarding that is not very realistic. However, rendering 294,751,494 elements (as in the BenchmarkRaw/raw_element benchmark cited above) to a single `io.Writer` is probably just as unrealistic.

The `maragu.dev/gomponents/html` tests on maragudk/main are probably the "most real world" of any of the tests, as they render a whole page with a realistic number of elements and attributes to a buffer, which is then reset once for each page render. However, as I said above, 1. rendering to a `string.Builder` is not that realistic anyway, 2. I agree that we largely measure noise, and 3. without a reference, the value of the results is questionable anyway.

If we do decide that "real world" benchmarks are useful to have, I think we should rethink our approach to that, and use something actually realistic, like maybe `httptest.ResponseRecorder`. Anyway, I think that is somewhat orthogonal to the issue at hand, and should be discussed separately.

## My Suggestion

My alternate suggestion for getting rid of the benchmarking noise would be to [render to `io.Discard`](https://github.com/Hendrikto/gomponents/tree/dev-milanobrtlik-benchmark), which is the idiomatic Go way of discarding output. We never actually inspect any of the buffers after writing to them. The correct **behavior** is verified in other tests, and the benchmarks are not concerned with that, but rather just with **performance**.

As @milanobrtlik said, this PR is a half-measure, getting rid of most but not all of the noise. I think if we are at it, we should eliminate the noise fully. We just need something to render to, but we do not really care what that is, and we never look at the output. `io.Discard` exists for exactly that use case.

On my branch implementing this idea, I get the following results:

<details>
<summary>hendrikto/dev-milanobrtlik-benchmark</summary>

```
❯ git rev-parse HEAD
318b6b5c0946ed57b144aa63103a4faa67b1d5e7 # hendrikto/dev-milanobrtlik-benchmark
goos: linux
goarch: amd64
pkg: maragu.dev/gomponents
BenchmarkAttr/boolean_attributes-32         	100000000	       11.39 ns/op	     24 B/op	      1 allocs/op
BenchmarkAttr/name-value_attributes-32      	39067375	       32.04 ns/op	     48 B/op	      1 allocs/op
BenchmarkEl/normal_elements-32              	68803082	       16.86 ns/op	     48 B/op	      1 allocs/op
BenchmarkRaw/raw_element-32                 	438959364	        2.734 ns/op	      0 B/op	      0 allocs/op
BenchmarkRawf/formatted_raw_element-32      	28177494	       41.75 ns/op	     40 B/op	      2 allocs/op
BenchmarkText/simple_text_element-32        	66672211	       19.17 ns/op	     16 B/op	      1 allocs/op
BenchmarkTextf/formatted_text_element-32    	24354804	       49.67 ns/op	     40 B/op	      2 allocs/op
PASS
ok  	maragu.dev/gomponents	8.466s
PASS
ok  	maragu.dev/gomponents/components	0.002s
goos: linux
goarch: amd64
BenchmarkRealisticPage/construct_and_render-32         	  14258	    83832 ns/op	 129964 B/op	   3048 allocs/op
BenchmarkRealisticPage/render_pre-built_tree-32        	  26853	    44711 ns/op	   1424 B/op	     19 allocs/op
PASS
ok  	maragu.dev/gomponents/html	2.400s
PASS
ok  	maragu.dev/gomponents/http	0.002s
?   	maragu.dev/gomponents/internal/assert	[no test files]
PASS
ok  	maragu.dev/gomponents/internal/import	0.002s
PASS
ok  	maragu.dev/gomponents/x/slices	0.002s
go test ./... -bench=. -benchmem  12.74s user 0.60s system 121% cpu 11.009 total
```
</details>

The core results are the same, but we also get rid of the last bit of remaining benchmark harness noise. This also makes the code simpler, easier to read, and more idiomatic. Furthermore, it makes the benchmarks execute faster, which is useful in itself (as long as the results stay meaningful, of course).

## Summary and Conclusions

* I agree that it is useful to get rid of the benchmark noise.
* I suggest a different approach than @milanobrtlik, specifically [rendering to `io.Discard`](https://github.com/Hendrikto/gomponents/tree/dev-milanobrtlik-benchmark).
* I recommend that approach unconditionally for the `maragu.dev/gomponents` benchmarks.
* For the `maragu.dev/gomponents/html` benchmarks it is less clear how to proceed, but I am leaning towards applying the noise reduction approach there as well.
* If we want something more closely resembling real world usage, we should rethink that separately, and probably use something like `httptest.ResponseRecorder`.
* As @milanobrtlik pointed out, the current benchmarks are pretty far from realistic anyway. Moreover, we lack reference values to compare to.

## Future Work

* I think we should introduce some benchmarks measuring whether using `WriteString` is actually worth it. It does make the code more indirect and complicated, and should therefore be justified. We currently lack any reference value to compare to.
* If we decide that realistic benchmarks are useful and something we want to have, we should actually make them realistic. They currently are not.

milanobrtlik:
Thanks, @Hendrikto — that's a thorough write-up, and it changes my mind. Let's go with `io.Discard`.

I reproduced your results on my side, and then ran one extra experiment, because I wanted to know whether the choice between our two approaches actually matters. Setup: all three variants on the same base (`ff049a4`), go1.24.10, linux/amd64, `-count=8`, compared with `benchstat`.

**Allocations are identical between `bytes.Buffer` and `io.Discard`.** Row for row, all nine benchmarks — the only difference anywhere is 1819 B/op vs 1812 B/op on `render pre-built tree`. So on the axis this PR was about, the two approaches are a tie rather than a half-measure against a complete one. The residual noise you're removing is entirely in ns/op: it is the memcpy into the buffer. That makes the question purely whether writing bytes somewhere is part of what we want to measure, and as I said, I don't hold that view strongly. Your version is simpler, faster and more idiomatic, so it should win.

**The stronger argument for the premise turns out not to be the `B/op` table I opened with.** I ran the escape fast path from #348 — a real library change — through all three harnesses. The current `strings.Builder` harness doesn't merely inflate the numbers, it reports statistically significant results that are not there:

```
Attr/boolean_attributes    33.20n → 34.03n   +2.48%  (p=0.026, n=8)
El/normal_elements         67.33n → 64.34n   -4.44%  (p=0.015, n=8)
```

Neither of those benchmarks escapes anything — a boolean attribute has no value, and `El("div")` has no text. Both numbers are harness noise that passed the significance test, in opposite directions. Under `bytes.Buffer` and under `io.Discard`, both rows correctly come out as `~` (p=0.49/0.51 and p=0.88/0.46). On the rows where #348 genuinely does something, all three harnesses agree on direction, and `bytes.Buffer` and `io.Discard` agree closely on magnitude (geomean -11.1% vs -12.2%).

I think that justifies the change better than "the numbers are too large" does: as it stands, the noise is capable of inventing a regression in code the change never touches.

**On your `WriteString` point:** `io.Discard` does satisfy `io.StringWriter`, so your branch still takes the same code path as before and hides nothing. But that also means it won't give you the reference value you're after — no harness we've discussed here ever takes the `Write([]byte)` branch. Measuring that needs a writer that deliberately doesn't implement `io.StringWriter`. Agreed it belongs in a separate change.

**One limitation worth recording, not an objection:** `io.Discard`'s `WriteString` is a no-op, so ns/op stops being sensitive to the *volume* of output. A change that emitted noticeably more bytes for the same number of calls would barely register. Output size is a library concern — escaping, attribute joining — so it is a real blind spot, just a narrow one, and ns/op is the noisiest metric we have anyway. I'd rather note it than trade away the simplicity for it.

**A small discrepancy for the record:** on go1.24.10, `BenchmarkRaw/raw_element` stays at 1 alloc/op, 16 B/op after the fix, rather than dropping to 0 as in your run. That's the go1.27 allocation work, not the harness — worth knowing before someone reads 0 allocs/op as a property of the library.

How would you like to proceed? Either works for me: I can repoint this PR at `io.Discard` and credit you, or close it in favour of a PR from your branch, which is arguably cleaner since both the idea and the implementation are yours. @markuswustenberg, whichever means less work for you.

markuswustenberg:
Thank you both for your inputs! @Hendrikto, greatly appreciated that you came back with a thorough second opinion.

My thoughts:
- I think it would make most sense to use io.Discard for the per-attribute and per-element tests. No noise, clean numbers, easy to reason about.
- I think for the complex documents, it would make sense to just have two sub-benchmarks measuring the same thing, but one with `io.Discard` and the other with a `bufio.Writer` or similar, depending on what the HTTP server actually uses underneath when writing out the response. That would get us the best of both worlds, IMO.
- Not sure whether it makes sense to throw away the `io.StringWriter` optimization, I'll defer that to later.

What do you think of that?

Hendrikto:
> it would make most sense to use io.Discard for the per-attribute and per-element tests. No noise, clean numbers, easy to reason about.

I agree.

> for the complex documents, it would make sense to just have two sub-benchmarks measuring the same thing, but one with io.Discard and the other with a bufio.Writer or similar

Makes sense.

> no harness we've discussed here ever takes the Write([]byte) branch. Measuring that needs a writer that deliberately doesn't implement io.StringWriter.

Yes, that was my point.

>  Agreed it belongs in a separate change.

> Not sure whether it makes sense to throw away the io.StringWriter optimization, I'll defer that to later.

I share that opinion. Specifically, 1. it is unclear whether that optimization is worth it, 2. we can deal with that later. I think this is orthogonal to this issue, just something that supported my reasoning.

>  I can repoint this PR at io.Discard and credit you, or close it in favour of a PR from your branch, which is arguably cleaner since both the idea and the implementation are yours.

Both sound fine to me. Either way, I think credit should be shared. I could add Co-authored-by or something similar, or you could do that too.

milanobrtlik:
Pushed, as a single commit replacing the old one: @Hendrikto's `io.Discard` for the per-attribute and per-element benchmarks, and @markuswustenberg's writer axis on `BenchmarkRealisticPage`. Co-authored with @Hendrikto.

The `net/http` breakdown above matches what I had found independently on go1.24.10 — same chain, same 2048, only the line numbers shifted by the version — so `bufio.NewWriterSize(io.Discard, 2048)` with a `Flush()` per iteration is what landed.

```
construct and render/discarded     126.2µs   127.3KiB   3058 allocs/op
construct and render/buffered      145.5µs   127.3KiB   3058 allocs/op
render pre-built tree/discarded    53.04µs   1.812KiB     28 allocs/op
render pre-built tree/buffered     69.10µs   1.812KiB     28 allocs/op
```

`-count=8`, go1.24.10, linux/amd64. Allocations are identical between the two writers, so the entire difference is the cost of delivering the bytes: 15% of construct-and-render, but 30% of rendering a pre-built tree. That the share moves that much between the two rows seems worth having on its own.

I also checked that the `buffered` variants don't quietly reintroduce the problem. Against #348 they report `Attr/boolean_attributes` at p=1.000 and `El/normal_elements` at p=0.959, and they see the real improvement in the same direction as `discarded`, just diluted by the writer's fixed cost (-9.4% against -11.5%). Which is the honest number a handler would actually see.

**One question on scope.** @markuswustenberg, I read "two sub-benchmarks measuring the same thing" as adding the writer axis on top of the existing construct/pre-built split rather than replacing it, since replacing it would lose the pre-building comparison that @Hendrikto singled out as newly visible. That gives four sub-benchmarks. If you meant two in total, say so and I'll cut it back.

**One finding worth recording, which qualifies something I claimed earlier.** `Rawf/formatted_raw_element` moves -4.50% (p=0.006, n=12) under #348 on the clean harness — and `Rawf` is `raw(fmt.Sprintf(...))`, which escapes nothing at all. It reproduces, so it isn't noise. The cause is code layout: before #348, `template.HTMLEscapeString` was inlined at three call sites; after, it is inlined once inside `escapeString`, which is not itself inlinable. The package's text grows by 94 bytes and every function after it moves.

So the clean harnesses remove the harness's own allocator noise, which is what produced the two false positives I posted earlier — those still come out as `~` — but they cannot remove code layout effects, and nothing can. Worth knowing before reading the table in #348, where `Rawf` will look like an improvement that has nothing to do with the change.

Hendrikto (approved):

markuswustenberg:
@milanobrtlik this additional subtests is exactly what I meant, thank you.

Thank you both! Looks like this is ready to merge.

PR #347, opened 2026-09-01
Reduce allocations in JoinAttrs

milanobrtlik:
Rebased onto #346, now that it's on main. The benchmark added here renders to `io.Discard`, following the convention settled there — without it, it would measure its own output buffer re-growing more than it measures `JoinAttrs`.

Three commits, each standing on its own:

**1. Add a `JoinAttrs` benchmark.** There wasn't one. It covers a flat child list, the nested groups #335 taught it to flatten, and a list with nothing to join. Construct is split from construct-and-render, because `JoinAttrs` does all of its work when it's called — render turns out to add ~10ns and no allocations at all.

**2. Hoist the prefixes and give the slices their capacity.** `extractAttrValue` rebuilt `" "+name` and `" "+name+`="`` for every child it inspected, though `name` is fixed for the whole call. It was the largest CPU entry after the render itself, showing up as `runtime.concatstrings`. `attrValues` stays lazily allocated so a child list with no matching attribute doesn't pay for it.

**3. Reuse one buffer.** Every child got its own `strings.Builder`, re-grown from nil each time — about 78% of all remaining allocations. One `bytes.Buffer` in `JoinAttrs` instead, whose `Reset` keeps its capacity, matching on bytes so the per-child string goes away too.

Cumulative, `BenchmarkJoinAttrs`, `-count=10`, go1.24.10, linux/amd64:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| flat | -47% | -36% | 25 → 10 |
| nested groups | -47% | -44% | 25 → 10 |
| no match | -49% | -44% | 16 → 4 |

Rendered output is unchanged throughout.

Commit 3 turned up an edge case worth pinning down. A node that renders as just `` name="`` ends with the same quote the prefix opens the value with, so slicing the value out naively runs off the end. It has always been treated as an attribute with no value, and still is — there are now tests for it and for the case where a real value follows it. Those tests pass against the code as it is on main too, so they document existing behaviour rather than new behaviour.

One thing I deliberately left alone: `extractAttrValue` renders each attribute, which escapes the value, then unescapes it, and the joined attribute escapes it again on render. `template.HTMLEscapeString` is about 13% of CPU purely because of that round trip. Avoiding it means getting at an attribute's value without rendering it, which needs something new in the root package — a bigger design question than this PR, and your call rather than mine.

### DIFF
--- a/components/benchmark_test.go
+++ b/components/benchmark_test.go
@@ -1,0 +1,69 @@
+//go:build go1.24
+
+package components_test
+
+import (
+	"io"
+	"testing"
+
+	g "maragu.dev/gomponents"
+	. "maragu.dev/gomponents/components"
+	. "maragu.dev/gomponents/html"
+)
+
+func BenchmarkJoinAttrs(b *testing.B) {
+	flat := func() []g.Node {
+		return []g.Node{
+			Class("inline-flex items-center"),
+			ID("badge"),
+			Class("px-2.5 py-0.5 rounded-full"),
+			Type("button"),
+			Class("bg-green-100 text-green-800"),
+		}
+	}
+
+	nested := func() []g.Node {
+		return []g.Node{
+			Class("inline-flex items-center"),
+			g.Group{
+				ID("badge"),
+				g.Group{
+					Class("px-2.5 py-0.5 rounded-full"),
+					Class("bg-green-100 text-green-800"),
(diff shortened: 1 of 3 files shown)


markuswustenberg:
Thanks!

PR #348, opened 2026-09-01
Skip rebuilding an ASCII set on every escape

milanobrtlik:
`template.HTMLEscapeString` decides whether a string needs escaping with `strings.ContainsAny(s, "'\"&<>\000")`, and `strings.IndexAny` builds a 32-byte ASCII bitmap out of that cutset **on every call**. gomponents calls it for every attribute value and every text node, so the bitmap gets rebuilt thousands of times per page for a set of six characters that never changes.

This checks those six against a table built once, and hands anything that actually needs escaping to `template.HTMLEscapeString` unchanged. Strings with nothing to escape are returned as they are — which is what `template.HTMLEscapeString` already did for them.

`BenchmarkRealisticPage`, `-count=10`, go1.24.10, linux/amd64, against `2b86aed`:

| | ns/op |
|---|---|
| construct and render, discarded | -7.9% |
| construct and render, buffered | -9.7% |
| render pre-built tree, discarded | -11.0% |
| render pre-built tree, buffered | -10.6% |

These replace an earlier `-12.1%` / `-10.6%` measured before #346 landed. The old harness re-grew its output buffer inside the baseline, which inflated the construct-and-render figure; this is the same effect #346 was about, and it turns out to have been flattering this PR.

Allocations are unchanged; this is purely the escaping decision. In the CPU profile `template.HTMLEscapeString` went from 26% of a page render to 11%, and what's left is the scan itself, which can't be avoided — you have to look at every byte to know there's nothing to escape.

**On testing.** This decides *whether* escaping happens, so a false negative would pass markup straight through. The tests check it against `template.HTMLEscapeString` for every single byte, every one of the 65536 pairs of bytes, every byte embedded in a string long enough to take `strings.IndexAny`'s other branch, and a set of named cases. There's also a fuzz target; 31 million executions found no difference. I'd rather over-test this one than argue it's obviously correct.

Two things I tried and threw away, in case they come up:

- Hoisting `isVoidElement` out of the `El` render closure. The captured bool pushes the closure into the next size class — `BenchmarkEl` goes 48 to 64 B/op, and `RealisticPage` construct-and-render costs 10% more memory for no time saved. Only the pre-built-tree case gains, and that's the rarer one.
- Keeping the six characters in a 64-bit constant instead of the table, since they're all below 0x40. 8.9% slower on `RealisticPage`. The table stays in L1 and an indexed load beats a compare plus a shift.


Hendrikto:
I also noticed the potential inefficiency with `template.HTMLEscapeString` before, and experimented with replacing it with `html.EscapeString`, which results in a small performance increase.

That work then stalled, as the effects were smaller than anticipated, and I had other stuff going on. Anyway, seeing this PR reminded me of that work, and I got the idea of combining both approaches.

The behavior is very slightly different. `template.HTMLEscapeString` escapes `\000` while `html.EscapeString` does not. I do not think that it matters to us, does it? If it does, it would be trivial to construct our own string replacer analogous to `html.htmlEscaper`, exactly matching the behavior of `template.HTMLEscapeString`.

Disappointingly, it did not really change much, and the numbers roughly match up with your branch. There might be a slight advantage of using `html.EscapeString`, but the delta is pretty small, making it hard to measure consistently. It would make sense logically, as `html.EscapeString` itself is faster than `template.HTMLEscapeString`. Nonetheless, I think it is useful to have additional data on this.

I took the milanobrtlik/benchmark-buffer  (f1d6641d61b9c19da6cacd192230f343ad5c64ac) as the baseline, and rebased everything against that.

<details>
<summary>full benchmarking details</summary>

```
goos: linux
goarch: amd64
pkg: maragu.dev/gomponents
BenchmarkAttr/boolean_attributes-32         	100000000	       11.35 ns/op	     24 B/op	      1 allocs/op
BenchmarkAttr/name-value_attributes-32      	36927346	       32.57 ns/op	     48 B/op	      1 allocs/op
BenchmarkEl/normal_elements-32              	64281360	       18.71 ns/op	     48 B/op	      1 allocs/op
BenchmarkRaw/raw_element-32                 	425137560	        2.788 ns/op	      0 B/op	      0 allocs/op
BenchmarkRawf/formatted_raw_element-32      	27347894	       43.63 ns/op	     40 B/op	      2 allocs/op
BenchmarkText/simple_text_element-32        	61790970	       19.72 ns/op	     16 B/op	      1 allocs/op
BenchmarkTextf/formatted_text_element-32    	23800184	       51.27 ns/op	     40 B/op	      2 allocs/op
PASS
ok  	maragu.dev/gomponents	8.364s
PASS
ok  	maragu.dev/gomponents/components	0.002s
goos: linux
goarch: amd64
BenchmarkRealisticPage/construct_and_render/discarded-32         	  13760	    87025 ns/op	 129967 B/op	   3048 allocs/op
BenchmarkRealisticPage/construct_and_render/buffered-32          	  11635	   103236 ns/op	 129965 B/op	   3048 allocs/op
BenchmarkRealisticPage/render_pre-built_tree/discarded-32        	  24746	    47817 ns/op	   1424 B/op	     19 allocs/op
BenchmarkRealisticPage/render_pre-built_tree/buffered-32         	  19017	    61255 ns/op	   1424 B/op	     19 allocs/op
PASS
ok  	maragu.dev/gomponents/html	4.752s
PASS
ok  	maragu.dev/gomponents/http	0.002s
?   	maragu.dev/gomponents/internal/assert	[no test files]
PASS
ok  	maragu.dev/gomponents/internal/import	0.002s
PASS
ok  	maragu.dev/gomponents/x/slices	0.001s
go test ./... -bench=. -benchmem  15.36s user 0.61s system 120% cpu 13.254 total

❯ git ll main..
122f73f 2026-09-03 Milan Obrt.. Skip rebuilding an ASCII set on every escape  (HEAD -> escape-fast-path)
f1d6641 2026-09-03 Milan Obrt.. Benchmark against writers that don't allocate per render  (milanobrtlik/benchmark-buffer, benchmark-buffer)
goos: linux
goarch: amd64
pkg: maragu.dev/gomponents
BenchmarkAttr/boolean_attributes-32         	100000000	       11.69 ns/op	     24 B/op	      1 allocs/op
BenchmarkAttr/name-value_attributes-32      	58848516	       19.20 ns/op	     48 B/op	      1 allocs/op
BenchmarkEl/normal_elements-32              	62884564	       18.19 ns/op	     48 B/op	      1 allocs/op
BenchmarkRaw/raw_element-32                 	450394694	        2.645 ns/op	      0 B/op	      0 allocs/op
BenchmarkRawf/formatted_raw_element-32      	28650559	       42.34 ns/op	     40 B/op	      2 allocs/op
BenchmarkText/simple_text_element-32        	76425560	       15.81 ns/op	     16 B/op	      1 allocs/op
BenchmarkTextf/formatted_text_element-32    	25399047	       47.98 ns/op	     40 B/op	      2 allocs/op
PASS
ok  	maragu.dev/gomponents	8.287s
PASS
ok  	maragu.dev/gomponents/components	0.002s
goos: linux
goarch: amd64
BenchmarkRealisticPage/construct_and_render/discarded-32         	  15939	    74135 ns/op	 129966 B/op	   3048 allocs/op
BenchmarkRealisticPage/construct_and_render/buffered-32          	  13141	    91031 ns/op	 129967 B/op	   3048 allocs/op
BenchmarkRealisticPage/render_pre-built_tree/discarded-32        	  31250	    38007 ns/op	   1424 B/op	     19 allocs/op
BenchmarkRealisticPage/render_pre-built_tree/buffered-32         	  20923	    56609 ns/op	   1424 B/op	     19 allocs/op
PASS
ok  	maragu.dev/gomponents/html	4.754s
PASS
ok  	maragu.dev/gomponents/http	0.002s
?   	maragu.dev/gomponents/internal/assert	[no test files]
PASS
ok  	maragu.dev/gomponents/internal/import	0.002s
PASS
ok  	maragu.dev/gomponents/x/slices	0.001s
go test ./... -bench=. -benchmem  15.36s user 0.66s system 121% cpu 13.182 total

❯ git ll main..
33098ea 2026-09-03 Hendrikto    Use html.EscapeString  (HEAD -> html.EscapeString, hendrik/html.EscapeString)
f1d6641 2026-09-03 Milan Obrt.. Benchmark against writers that don't allocate per render  (milanobrtlik/benchmark-buffer, benchmark-buffer)
goos: linux
goarch: amd64
pkg: maragu.dev/gomponents
BenchmarkAttr/boolean_attributes-32         	100000000	       11.63 ns/op	     24 B/op	      1 allocs/op
BenchmarkAttr/name-value_attributes-32      	50310184	       20.97 ns/op	     48 B/op	      1 allocs/op
BenchmarkEl/normal_elements-32              	66392806	       18.20 ns/op	     48 B/op	      1 allocs/op
BenchmarkRaw/raw_element-32                 	463853727	        2.578 ns/op	      0 B/op	      0 allocs/op
BenchmarkRawf/formatted_raw_element-32      	27038228	       44.97 ns/op	     40 B/op	      2 allocs/op
BenchmarkText/simple_text_element-32        	59091823	       19.36 ns/op	     16 B/op	      1 allocs/op
BenchmarkTextf/formatted_text_element-32    	21810418	       52.08 ns/op	     40 B/op	      2 allocs/op
PASS
ok  	maragu.dev/gomponents	8.124s
PASS
ok  	maragu.dev/gomponents/components	0.002s
goos: linux
goarch: amd64
BenchmarkRealisticPage/construct_and_render/discarded-32         	  14301	    83177 ns/op	 129859 B/op	   3045 allocs/op
BenchmarkRealisticPage/construct_and_render/buffered-32          	  12147	    99082 ns/op	 129860 B/op	   3045 allocs/op
BenchmarkRealisticPage/render_pre-built_tree/discarded-32        	  25483	    46339 ns/op	   1424 B/op	     19 allocs/op
BenchmarkRealisticPage/render_pre-built_tree/buffered-32         	  18830	    63136 ns/op	   1424 B/op	     19 allocs/op
PASS
ok  	maragu.dev/gomponents/html	4.767s
PASS
ok  	maragu.dev/gomponents/http	0.002s
?   	maragu.dev/gomponents/internal/assert	[no test files]
PASS
ok  	maragu.dev/gomponents/internal/import	0.002s
PASS
ok  	maragu.dev/gomponents/x/slices	0.002s
go test ./... -bench=. -benchmem  15.10s user 0.59s system 120% cpu 13.025 total

❯ git ll main..
920e7a6 2026-09-03 Hendrikto    Use html.EscapeString  (HEAD -> escape-fast-path_with_html.EscapeString)
122f73f 2026-09-03 Milan Obrt.. Skip rebuilding an ASCII set on every escape  (escape-fast-path)
f1d6641 2026-09-03 Milan Obrt.. Benchmark against writers that don't allocate per render  (milanobrtlik/benchmark-buffer, benchmark-buffer)
goos: linux
goarch: amd64
pkg: maragu.dev/gomponents
BenchmarkAttr/boolean_attributes-32         	100000000	       10.47 ns/op	     24 B/op	      1 allocs/op
BenchmarkAttr/name-value_attributes-32      	70133464	       16.80 ns/op	     48 B/op	      1 allocs/op
BenchmarkEl/normal_elements-32              	71062383	       16.77 ns/op	     48 B/op	      1 allocs/op
BenchmarkRaw/raw_element-32                 	437179575	        2.745 ns/op	      0 B/op	      0 allocs/op
BenchmarkRawf/formatted_raw_element-32      	28047568	       41.66 ns/op	     40 B/op	      2 allocs/op
BenchmarkText/simple_text_element-32        	73675819	       15.45 ns/op	     16 B/op	      1 allocs/op
BenchmarkTextf/formatted_text_element-32    	25726011	       48.00 ns/op	     40 B/op	      2 allocs/op
PASS
ok  	maragu.dev/gomponents	8.170s
PASS
ok  	maragu.dev/gomponents/components	0.002s
goos: linux
goarch: amd64
BenchmarkRealisticPage/construct_and_render/discarded-32         	  16076	    74491 ns/op	 129860 B/op	   3045 allocs/op
BenchmarkRealisticPage/construct_and_render/buffered-32          	  13461	    88980 ns/op	 129860 B/op	   3045 allocs/op
BenchmarkRealisticPage/render_pre-built_tree/discarded-32        	  30808	    38739 ns/op	   1424 B/op	     19 allocs/op
BenchmarkRealisticPage/render_pre-built_tree/buffered-32         	  21056	    55233 ns/op	   1424 B/op	     19 allocs/op
PASS
ok  	maragu.dev/gomponents/html	4.757s
PASS
ok  	maragu.dev/gomponents/http	0.002s
?   	maragu.dev/gomponents/internal/assert	[no test files]
PASS
ok  	maragu.dev/gomponents/internal/import	0.002s
PASS
ok  	maragu.dev/gomponents/x/slices	0.001s
go test ./... -bench=. -benchmem  14.92s user 0.66s system 119% cpu 13.057 total
```
</details>

| benchmark                                              | baseline (milanobrtlik/benchmark-buffer) | milanobrtlik/escape-fast-path | hendrik/html.EscapeString | hendrik/escape-fast-path_with_html.EscapeString |
|--------------------------------------------------------|------------------------------------------|-------------------------------|---------------------------|-------------------------------------------------|
| BenchmarkAttr/boolean_attributes                       |     11.35  ns/op                         |    11.69  ns/op               |    11.63  ns/op           |    10.47  ns/op                                 |
| BenchmarkAttr/name-value_attributes                    |     32.57  ns/op                         |    19.20  ns/op               |    20.97  ns/op           |    16.80  ns/op                                 |
| BenchmarkEl/normal_elements                            |     18.71  ns/op                         |    18.19  ns/op               |    18.20  ns/op           |    16.77  ns/op                                 |
| BenchmarkRaw/raw_element                               |      2.788 ns/op                         |     2.645 ns/op               |     2.578 ns/op           |     2.745 ns/op                                 |
| BenchmarkRawf/formatted_raw_element                    |     43.63  ns/op                         |    42.34  ns/op               |    44.97  ns/op           |    41.66  ns/op                                 |
| BenchmarkText/simple_text_element                      |     19.72  ns/op                         |    15.81  ns/op               |    19.36  ns/op           |    15.45  ns/op                                 |
| BenchmarkTextf/formatted_text_element                  |     51.27  ns/op                         |    47.98  ns/op               |    52.08  ns/op           |    48.00  ns/op                                 |
| BenchmarkRealisticPage/construct_and_render/discarded  |  87025     ns/op                         | 74135     ns/op               | 83177     ns/op           | 74491     ns/op                                 |
| BenchmarkRealisticPage/construct_and_render/buffered   | 103236     ns/op                         | 91031     ns/op               | 99082     ns/op           | 88980     ns/op                                 |
| BenchmarkRealisticPage/render_pre-built_tree/discarded |  47817     ns/op                         | 38007     ns/op               | 46339     ns/op           | 38739     ns/op                                 |
| BenchmarkRealisticPage/render_pre-built_tree/buffered  |  61255     ns/op                         | 56609     ns/op               | 63136     ns/op           | 55233     ns/op                                 |

In summary, I can reproduce the performance advantage of this PR, which is superior to my idea of replacing `template.HTMLEscapeString` with `html.EscapeString`. Combining both approaches might have a slight advantage, but it is hard to measure.

markuswustenberg:
Would this kind of performance optimization be worth doing upstream, in the Go stdlib instead? Seems weird we have to have our own string escaping, but maybe if we make it only temporarily?

Hendrikto:
I thought about this as well. It is also pretty weird to me that there are two slightly different HTML escaping functions in the standard library. They will probably have to be kept, including the diverging behavior, for backwards-compatibility, but maybe we could try to upstream the performance improvements.

markuswustenberg:
No matter what it could be worth it doing here, so Go versions without any potential future changes get the optimization as well. Then I'll leave it up to you whether you want to take it upstream as well. :-)

So: this change is basically good to go as-is, right?

milanobrtlik:
Almost — one thing first, and thanks for updating the branch.

The table in the description was measured before #346 landed, on the harness that #346 replaced, so it was overstating things. I've updated it: re-measured against `2b86aed` with `-count=10`, construct-and-render comes out at **-7.9%** rather than `-12.1%`, because the old harness kept its own buffer re-growth in the baseline. Pre-built tree is unchanged at `-11.0%`. The description now carries the full four rows and the figures it replaces, so the correction is on the record rather than quietly swapped in. Allocations are still unchanged and the testing story is the same.

Which makes this the first thing measured honestly by the change it prompted, and it comes out smaller. Fair enough.

**On taking it upstream.** I dug into where the cost actually is, and I think the target is narrower than "HTML escaping". `HTMLEscapeString` calls `strings.ContainsAny(s, "'\"&<>\000")`, which is `IndexAny`, which calls `makeASCIISet(chars)` on every call. So the waste isn't in `text/template` at all — it's that `strings.IndexAny` can't hoist the set for a cutset the caller passes as a constant. Fixing it there would help every caller with a fixed cutset, not just this one, which seems like the more useful proposal.

But I don't think that would fully subsume this change, for a reason worth knowing either way: the ASCII-set path is gated on `len(s) > 8`. Anything shorter falls through to `for i, c := range s { IndexRune(chars, c) }`, which rescans the cutset for every byte. Attribute values and short text nodes land in that branch constantly. A stdlib fix that only hoists the set would leave them exactly as they are, while the table here covers both branches — that's why the tests deliberately include strings on both sides of the length-8 boundary.

So I'd say it stands on its own rather than as a placeholder, and @Hendrikto, upstreaming is yours if you want it — you found the `strings` angle first.

**@Hendrikto, on `html.EscapeString`:** you're right that NUL is the difference, and it's the *only* one. I compared the two over every byte and then over every rune in `0..0x10FFFF`; `\000` is the single divergence in the entire input space.

But I don't think it needs deciding here, because combining the two doesn't buy anything measurable. `-count=10` with benchstat, your combination against this PR: all eleven rows come out `~`, geomean -0.25%. That follows from the fast path — the delegate only runs for strings that actually contain one of the six characters, and in `RealisticPage` exactly one string out of 235 does, while none of the micro benchmarks do. So swapping the delegate has almost nothing to act on. I think your table is picking up code layout rather than the escaper, which would also explain the opposite signs across the four `RealisticPage` rows.

That said, your instinct that `html/template` is the wrong dependency for this is right, and there's a better argument for it than speed that I hadn't seen until I went looking. I'll open it separately rather than load it onto this one.

PR #349, opened 2026-09-08
Avoid an allocation and a nil-sibling panic in Doctype

milanobrtlik:
Two independent fixes in `Doctype`, one small commit each.

**1. Use `io.WriteString`.** `w.Write([]byte("<!doctype html>"))` — `w` is the interface parameter of a closure, so the conversion can't be devirtualized and escapes:

```
html/elements.go:17:31: ([]byte)("<!doctype html>") escapes to heap
```

That's one allocation on every render of every page built with `components.HTML5`. cca0e29 (#322) moved the root package to `io.WriteString` but didn't reach this file, so it's a leftover from that refactor. `io.WriteString` takes the `WriteString` fast path on the writers this meets in practice (`http.ResponseWriter`, `bufio.Writer`, `strings.Builder`) and falls back to exactly the old behaviour on anything else.

A pre-built `Doctype(g.El("html"))` rendered `-count=10`, go1.24.10, linux/amd64, through both writers, since `net/http` puts a 2 KiB `bufio.Writer` in front of a handler's `ResponseWriter`:

| writer | before | after | |
|---|---|---|---|
| `io.Discard` | 24.36n ± 5% | 11.34n ± 2% | -53.46% (p=0.000 n=10) |
| 2 KiB `bufio.Writer` | 34.52n ± 3% | 20.46n ± 2% | -40.72% (p=0.000 n=10) |

`B/op` goes 16 to 0 and `allocs/op` 1 to 0 on both.

Those are the numbers for the doctype on its own. In a page they are one allocation out of sixteen: `Doctype(HTML(Head(TitleEl(...)), Body(P(...))))` through the buffer keeps the allocation saving and puts the time difference in the noise (p=0.529).

**2. Ignore a nil sibling.** `Doctype` renders its sibling unconditionally, so a nil one panics:

```go
html.Doctype(g.If(false, html.HTML())).Render(w)
// panic: runtime error: invalid memory address or nil pointer dereference
```

`g.If` returns a genuine nil interface when its condition is false, and `Iff` does the same, so this comes from ordinary conditional markup rather than from passing a literal `nil`. The panic also lands *after* `<!doctype html>` has gone out, so under `http.Adapt` net/http's recover kills the connection mid-response — `Adapt`'s own nil guard doesn't help, because `Doctype(nil)` is itself a non-nil `NodeFunc`.

Everywhere else nil nodes are ignored (`renderChild` guards for it, AGENTS.md states it as a rule), and `Doctype` was the only place rendering a caller-supplied `Node` without the check. The doctype itself is still written when the sibling is nil — the sibling is the nil node, not the whole thing.

The only behaviour change is that `Doctype(nil)` now renders `<!doctype html>` instead of panicking. Coverage stays at 100%.


html/elements.go
```diff
@@ -12,11 +12,17 @@ import (
 )
 
 // Doctype returns a special kind of [g.Node] that prefixes its sibling with the string "<!doctype html>".
+// A nil sibling is ignored, like nil nodes are elsewhere in the library.
```
markuswustenberg: Maybe just "If sibling is nil, only the doctype is rendered."? No need to reference outside the package necessarily.

html/elements_test.go
```diff
@@ -21,6 +21,11 @@ func TestDoctype(t *testing.T) {
 		assert.Equal(t, `<!doctype html><html></html>`, Doctype(g.El("html")))
 	})
 
+	t.Run("ignores a nil sibling", func(t *testing.T) {
+		assert.Equal(t, `<!doctype html>`, Doctype(nil))
+		assert.Equal(t, `<!doctype html>`, Doctype(g.If(false, g.El("html"))))
```
markuswustenberg, line 26: Arguably unnecessary line that tests the same thing, but I get why you included it. Your choice.

markuswustenberg (approved):

html/elements.go
milanobrtlik: Done.

html/elements_test.go
milanobrtlik: Keeping it, it's the `g.If` case from the description.

PR #350, opened 2026-09-08
Drop the html/template dependency

milanobrtlik:
Following up on my last comment on #348, where I said I'd open this separately.

After the escape fast path landed, `template.HTMLEscapeString` was the only thing left in the library reaching into `html/template` — one six-character escaper in exchange for the heaviest dependency in the module. This replaces it with a `strings.NewReplacer` holding exactly the six replacements `text/template.HTMLEscape` makes.

**Output is unchanged, byte for byte.**

Seventeen packages leave the root package's dependency set — `text/template`, `text/template/parse`, `regexp`, `regexp/syntax`, `encoding/json`, `encoding/base64`, `encoding/binary`, `encoding`, `net/url`, `net/netip`, `path/filepath`, `unicode/utf16`, `unique`, `weak`, `maps`, `bytes` and `html/template` itself. `go list -deps .` goes from **76 to 57**.

| | main | this PR | |
|---|---|---|---|
| minimal program | 2 787 119 B | 1 956 627 B | -811 KiB |
| `net/http` + `encoding/json` + `regexp` | 8 920 166 B | 8 837 036 B | **-81 KiB (-0.93%)** |

The second row is the honest headline. Most of those packages come back the moment an application imports `net/http` or `regexp` on its own; what's left is `text/template` and friends, which nothing else pulls in. So it's -0.9% of a real binary, not -30%.

**On trusting it.** The tests still import `html/template` and still compare against `HTMLEscapeString` — every byte, all 65536 pairs of bytes, strings either side of the length-8 boundary, the named cases, and the fuzz target (62 million executions, no divergence). Test imports don't count towards the library's dependencies, so the equivalence stays proven while the package stops shipping. If you'd rather keep delegating to the standard library, that's a fair call and this is easy to close — but I don't think the proof gets weaker by moving the escaper in-tree, given it's six replacements checked against the reference on every input.

**Performance is unchanged**, as it has to be: the replacer only runs for strings that actually contain one of the six characters, which in `BenchmarkRealisticPage` is one string out of 235. Two independent `-count=10` rounds on the same two binaries give geomean `+1.4%` and `-1.9%` — the sign flips, so that's layout and machine noise rather than the change.

@Hendrikto spotted this in #348. He proposed `html.EscapeString`, which differs from `text/template` on exactly one input (NUL); a local replacer needs no decision about that, because nothing about the behaviour changes.

### DIFF
--- a/gomponents_benchmark_test.go
+++ b/gomponents_benchmark_test.go
@@ -10,10 +10,6 @@ import (
 	g "maragu.dev/gomponents"
 )
 
-// escapedProse is the kind of user-written text that takes the escaping path. English prose
-// contains apostrophes, which is enough on its own.
-var escapedProse = strings.Repeat("It's a paragraph of user-written text, with quotes & apostrophes in it. ", 40)
-
 func BenchmarkAttr(b *testing.B) {
 	b.Run("boolean attributes", func(b *testing.B) {
 		for b.Loop() {
@@ -80,8 +76,12 @@ func BenchmarkText(b *testing.B) {
 	})
 
 	b.Run("prose element needing escaping", func(b *testing.B) {
+		// English prose contains apostrophes, which on its own is enough to take the
+		// escaping path.
+		prose := strings.Repeat("It's a paragraph of user-written text, with quotes & apostrophes in it. ", 40)
+
 		for b.Loop() {
-			e := g.Text(escapedProse)
+			e := g.Text(prose)
 			_ = e.Render(io.Discard)
 		}
 	})
--- a/gomponents_benchmark_test.go
+++ b/gomponents_benchmark_test.go
@@ -4,10 +4,15 @@ package gomponents_test
 
 import (
 	"io"
+	"strings"
 	"testing"
 
 	g "maragu.dev/gomponents"
 )
+
+// escapedProse is the kind of user-written text that takes the escaping path. English prose
+// contains apostrophes, which is enough on its own.
+var escapedProse = strings.Repeat("It's a paragraph of user-written text, with quotes & apostrophes in it. ", 40)
 
 func BenchmarkAttr(b *testing.B) {
 	b.Run("boolean attributes", func(b *testing.B) {
@@ -20,6 +25,13 @@ func BenchmarkAttr(b *testing.B) {
 	b.Run("name-value attributes", func(b *testing.B) {
 		for b.Loop() {
 			a := g.Attr("hat", "party")
+			_ = a.Render(io.Discard)
+		}
+	})
+
+	b.Run("name-value attributes needing escaping", func(b *testing.B) {
+		for b.Loop() {
+			a := g.Attr("hat", `"party" & fun`)
 			_ = a.Render(io.Discard)
 		}
 	})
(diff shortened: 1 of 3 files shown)


markuswustenberg:
Am I understanding this correctly, that the only difference is that the binary potentially gets slightly smaller? If that is the case, I'd rather keep the dependency and not duplicate this functionality from the stdlib, since binary size will be totally dominated by other imports in a typical app IMO. But I'm open to having my mind changed if you feel strongly about this. :-)

milanobrtlik:
Fair question, and my description undersold it. "Performance is unchanged" came from `BenchmarkRealisticPage`, where one string out of 235 contains a character that needs escaping. The root benchmarks were no better: `"some simple text"` and `"party"` contain none of the six characters, so the escaper was never measured.

I've added benchmarks that do take the path. `benchstat`, `-count=10`, main vs this branch:

| | main | this PR | |
|---|---|---|---|
| attribute value | 132.3 ns, 152 B, 6 allocs | 73.5 ns, 112 B, 3 allocs | -44% |
| 36-byte sentence | 132.9 ns, 176 B, 6 allocs | 91.1 ns, 112 B, 3 allocs | -31% |
| 2.8 KiB of prose | 4.04 µs, 15.2 KiB, 13 allocs | 2.42 µs, 6.3 KiB, 3 allocs | -40% |

All p=0.000. Benchmarks with nothing to escape are unchanged.

`HTMLEscapeString` converts to `[]byte` and then writes chunk by chunk into a `strings.Builder`, so its allocations grow with the content. The replacer's patterns are all single bytes, so it sizes the output in one pass and allocates once.

There is a case where it loses, though not where @maragubot put it. `byteStringReplacer` does switch at `len(toReplace)*countCutOff`, 48 bytes for six characters. But at 48 bytes the replacer is still the faster of the two here; the crossover is around 256, and past it the penalty grows with length instead of staying at ten nanoseconds. One `&` in the whole string, median of 10:

| length | stdlib | replacer | |
|---|---|---|---|
| 64 B | 112.0 ns | 88.8 ns | -21% |
| 256 B | 236.0 ns | 235.2 ns | -0.3% |
| 4 KiB | 2.73 µs | 2.91 µs | +6.6% |
| 16 KiB | 10.07 µs | 11.41 µs | +13.3% |

So the trade is: faster for short strings and for anything with more than a sprinkling of escapes, slower for long text with almost none. Prose is denser than that and wins by 2x to 4x.

I've added the boundary cases too: 47, 48 and 49 bytes with each escaped character at the front, the back and the middle, plus a long sparse string, a long dense one, and one that is nothing but replacements. Both `Replace` paths are now pinned by fixed tests rather than only by fuzzing.

So it isn't only binary size. Still your call.

gomponents_benchmark_test.go
```diff
@@ -4,11 +4,16 @@ package gomponents_test
...
 	"io"
+	"strings"
 	"testing"
 
 	g "maragu.dev/gomponents"
 )
 
+// escapedProse is the kind of user-written text that takes the escaping path. English prose
+// contains apostrophes, which is enough on its own.
+var escapedProse = strings.Repeat("It's a paragraph of user-written text, with quotes & apostrophes in it. ", 40)
```
markuswustenberg: inline?

milanobrtlik: Inlined. `b.Loop` resets the timer on its first call, so the setup stays out of the measurement; the numbers are unchanged.

markuswustenberg:
Excellent, thank you!

PR #352, opened 2026-09-10
Avoid another allocation

Hendrikto:
Now that we have direct access to a `strings.Replacer`, we can avoid another allocation.

| BenchmarkAttr/name-value_attributes_needing_escaping | before | after |
|------------------------------------------------------|--------|-------|
| ns/op                                                | 41.33  | 24.37 |
| B/op                                                 | 80     | 48    |
| allocs/op                                            | 2      | 1     |


Hendrikto:
We could also do this:

```diff
diff --git a/gomponents.go b/gomponents.go
index c9d5298..fd789aa 100644
--- a/gomponents.go
+++ b/gomponents.go
@@ -245,6 +245,19 @@ func escapeString(s string) string {
 	return s
 }
 
+func escapeWriteString(w io.Writer, s string) error {
+	for i := 0; i < len(s); i++ {
+		if htmlEscapeSet[s[i]] {
+			_, err := htmlEscaper.WriteString(w, s)
+			return err
+		}
+	}
+
+	_, err := io.WriteString(w, s)
+	return err
+}
+
 // valueAttr creates a name-value attribute Node.
 func valueAttr(name, value string) Node {
 	return attrFunc(func(w io.Writer) error {
@@ -260,7 +273,7 @@ func valueAttr(name, value string) Node {
 			return err
 		}
 
-		if _, err := htmlEscaper.WriteString(w, value); err != nil {
+		if err := escapeWriteString(w, value); err != nil {
 			return err
 		}
 
 ```
 
 But it does not make a difference according to my testing.

markuswustenberg:
Wouldn't this make the cases that don't need escaping worse, since we're no longer using the optimization with `htmlEscapeSet`? (I'll have my bot have a look.)

Hendrikto:
> Wouldn't this make the cases that don't need escaping worse

I did think about that, which is why I suggested the helper function. But I could not measure a difference with the existing benchmarks, as confirmed by @maragubot. It is all in the ±2% range. 

Honestly, I cannot follow @maragubot from that point onwards. If you have benchmarks that show the difference you claim, show them. There is nothing to run or reproduce here.

> (*strings.Replacer).WriteString does a sync.Once

So does `(*strings.Replacer).Replace`.

>  an interface dispatch to byteStringReplacer.WriteString

So does `(*strings.Replacer).Replace`, except to `byteStringReplacer.Replace`.

> The old path was a [256]bool scan

The new path is a `[256][]byte` scan. I do not think that should be slower when just checking for nil.

I cannot really tell if/how much of the response is just hallucinated, to be honest.

I will test the suggestion anyway and get back to you.

Hendrikto:
I added the suggested helper function, but without the redundant double check for `w.(io.StringWriter)`. We already know that if we reach that path, `w` is not an `io.StringWriter`, so there is no need to check again.

The benchmark results are unsurprisingly the same, as with the `escapeWriteString` version above. With a writer implementing `io.StringWriter`, there is no measurable difference.

The existing benchmarks may be insufficient, as we discussed before, but then we should actually discuss something concrete. @maragubot, show me what you got.

markuswustenberg:
> I did think about that, which is why I suggested the helper function. But I could not measure a difference with the existing benchmarks, as confirmed by @maragubot. It is all in the ±2% range.

Sorry, of course, I was a bit too quick with my answer.

I guess I'm just confused. If having the `htmlEscapeSet` optimization doesn't matter too much, how did we get to numbers that showed us it was a good idea? I'll have to go back and have a look. (Too much stuff in my head at once these days.)

> I added the suggested helper function, but without the redundant double check for `w.(io.StringWriter)`. We already know that if we reach that path, `w` is not an `io.StringWriter`, so there is no need to check again.
> 
> The benchmark results are unsurprisingly the same, as with the `escapeWriteString` version above. With a writer implementing `io.StringWriter`, there is no measurable difference.
> 
> The existing benchmarks may be insufficient, as we discussed before, but then we should actually discuss something concrete. @maragubot, show me what you got.

Yes, I think this is the right answer. Let's add some benchmarks that properly show the thing we're improving first. Then maybe I won't be so confused. :D

I'll get back to you. Thank you again for your work!

markuswustenberg:
@Hendrikto I've merged in the benchmark changes from #354. That should give us a better picture of your improvements. :-)

markuswustenberg:
@Hendrikto BTW, if TL;DR: You just need an extra subtest to get coverage back to 100%, then I think this is ready to merge.

Hendrikto:
I rebased against main, verifying that the diff with the previous version you pushed was empty.

Also, I added the requested test, bringing test coverage back up to 100%, and actually verifying that attributes are rendered correctly to `io.Writer`s that are not `io.StringWriter`s.

Hendrikto:
I addressed the linting error.

markuswustenberg:
Looks like your test doesn’t actually test the missing line? You need to write a value that must be escaped I think.

Hendrikto:
Oh yeah, sorry, my bad. I looked at the wrong file 😶

markuswustenberg:
Excellent! Thank you again for your work! :D It’s a joy to work with you.

PR #354, opened 2026-09-14
Benchmark rendering attributes to different writers

maragubot:
- Rewrites `BenchmarkAttr` as one table over three axes: construction mode (`construct and render`, `render pre-built`), writer (`io.Discard`, a `bufio.Writer` sized like net/http's response buffer, and both of those behind a wrapper that only has a `Write` method, like a struct embedding `http.ResponseWriter` does), and value (a boolean attribute plus a grid of short and long values, 16 and 256 bytes, at three levels of escaping: none, little at about one byte in sixteen, and much at one in four). The values are generated by a `value(n, escapes)` helper from the same base text, so within a length only the number of escaped bytes changes.
- Every sub-benchmark builds its own writer before its loop, so a buffer never inherits the fill level a previous sub-benchmark left in it.
- 56 sub-benchmarks, named like `BenchmarkAttr/render_pre-built/buffered/short_value,_little_escaping`, so a run can be filtered to one mode or writer. The three old sub-benchmark names are gone.
- Adds an implementation diary at `docs/diary/2026-09-14-attribute-writer-benchmarks.md` covering how the benchmark got to this shape through review.

The old `BenchmarkAttr` only wrote to `io.Discard`, whose writes are free and which has `WriteString`. That hides two things an escaping change can affect: the cost of issuing many small writes, and the path taken by writers that only implement `Write`. The long clean value makes the scan itself the dominant cost, so a change to how the scan is done shows up on its own. The construction axis keeps the closure allocation in view, so a change that trades constructor size for render speed shows both sides in one table.

For a concrete example, this is what the `render pre-built` rows show for #352 at e4173ec against main, Apple M4, Go 1.27.1, interleaved, n=10. The `construct and render` rows show the same pattern with one more allocation for the closure on every row.

| writer / value | main | #352 | sec/op | allocs/op |
|---|---|---|---|---|
| discarded / boolean | 3.80n | 3.77n | ~ | 0 → 0 |
| discarded / short value, no escaping | 12.0n | 11.9n | ~ | 0 → 0 |
| discarded / long value, no escaping | 76.7n | 76.6n | ~ | 0 → 0 |
| discarded / short value, little escaping | 33.2n | 18.1n | -46% | 1 → 0 |
| discarded / long value, little escaping | 243n | 133n | -45% | 2 → 0 |
| discarded / short value, much escaping | 36.5n | 23.1n | -37% | 1 → 0 |
| discarded / long value, much escaping | 331n | 243n | -26% | 2 → 0 |
| buffered / boolean | 7.52n | 7.45n | ~ | 0 → 0 |
| buffered / short value, no escaping | 21.4n | 21.5n | ~ | 0 → 0 |
| buffered / long value, no escaping | 90.1n | 90.8n | ~ | 0 → 0 |
| buffered / short value, little escaping | 42.7n | 30.7n | -28% | 1 → 0 |
| buffered / long value, little escaping | 255n | 182n | -29% | 2 → 0 |
| buffered / short value, much escaping | 45.8n | 45.2n | ~ | 1 → 0 |
| buffered / long value, much escaping | 349n | 428n | **+22%** | 2 → 0 |
| write-only / all seven | | | ~ | unchanged |
| write-only buffered / all seven | | | ~ | unchanged |

The `buffered / long value, much escaping` row is the one real trade-off in #352: with one byte in four escaped, streaming through the replacer issues 129 writes into the buffer where main issued one, and that costs more than the allocation it saves.

The write-only rows moved by up to +19% between the two binaries, but a single-binary A/B of the two code paths shows them identical, so that is binary layout rather than the change; see the comment below for details.

See #352.


gomponents_benchmark_test.go
```diff
@@ -33,6 +34,53 @@ func BenchmarkAttr(b *testing.B) {
...
+
+// BenchmarkAttrRender renders pre-built name-value attributes to writers with different
+// write costs, with values that take different paths through escaping.
+func BenchmarkAttrRender(b *testing.B) {
+	values := []struct {
+		Name, Value string
+	}{
+		{Name: "no escaping", Value: "party"},
+		{Name: "needing escaping", Value: `"party" & fun`},
+		{Name: "JSON needing escaping", Value: `{"hat":"party","glitter":true,"guests":["you","me"]}`},
```
markuswustenberg: Why is this one special? Because it's longer than the line just above? Maybe don't use JSON as an example, name it after which characteristic we're benchmarking.

gomponents_benchmark_test.go
```diff
@@ -33,6 +34,53 @@ func BenchmarkAttr(b *testing.B) {
...
+		{Name: "prose needing escaping", Value: strings.Repeat("It's a title with quotes & apostrophes in it. ", 4)},
```
markuswustenberg: same

gomponents_benchmark_test.go
maragubot: Agreed. Renaming to "many characters needing escaping" with a synthetic value of the same shape, `strings.Repeat(`"hat" & `, 6)`.

gomponents_benchmark_test.go
maragubot: Agreed. Renaming to "long value with few characters needing escaping"; the prose value stays since that sparse pattern is the characteristic.

gomponents_benchmark_test.go
```diff
@@ -33,6 +34,54 @@ func BenchmarkAttr(b *testing.B) {
...
+	w io.Writer
+}
+
+func (w writeOnly) Write(p []byte) (int, error) {
+	return w.w.Write(p)
+}
+
+// BenchmarkAttrRender renders pre-built name-value attributes to writers with different
+// write costs, with values that take different paths through escaping.
+func BenchmarkAttrRender(b *testing.B) {
```
markuswustenberg: Maybe just a subtest/subbenchmark in BenchmarkAttr?

maragubot: Agreed. Folding it into `BenchmarkAttr` as `render pre-built/<writer>/<value>` sub-benchmarks; the three existing names stay as they are.

markuswustenberg:
@Hendrikto, @maragubot and I added some more benchmarks which show the effects of your #352 change. In summary, nice work of you! :D

gomponents_benchmark_test.go
```diff
@@ -3,34 +3,86 @@
...
+type writeOnly struct {
+	w io.Writer
+}
+
+func (w writeOnly) Write(p []byte) (int, error) {
+	return w.w.Write(p)
+}
+
+// node keeps a constructed [g.Node] reachable across benchmark iterations.
+var node g.Node
```
markuswustenberg: No package-level state please, keep tests independent

maragubot: Agreed. A local `var node g.Node` in the sub-benchmark, outside the loop, keeps the same 48 B and 1 allocation; the package-level variable goes.

gomponents_benchmark_test.go
```diff
@@ -3,34 +3,85 @@
...
+		Name    string
+		Boolean bool
+		Value   string
+	}{
+		{Name: "boolean", Boolean: true},
+		{Name: "no escaping", Value: "party"},
+		{Name: "long value needing no escaping", Value: strings.Repeat("a title with no quotes or apostrophes in it ", 4)},
+		{Name: "needing escaping", Value: `"party" & fun`},
+		{Name: "many characters needing escaping", Value: strings.Repeat(`"hat" & `, 6)},
+		{Name: "long value with few characters needing escaping", Value: strings.Repeat("It's a title with quotes & apostrophes in it. ", 4)},
```
markuswustenberg: Streamline these?

- Short value no escaping
- Long value no escaping
- Short value little escaping
- Long value little escaping
- Short value much escaping
- Long value much escaping

maragubot: Agreed. Making it the full short/long by none/little/much grid, with the long value of each escaping level repeating the short pattern so only length changes.

PR #355, opened 2026-09-15
Automate coverage report generation

Hendrikto:
While checking test coverage for https://github.com/maragudk/gomponents/pull/352, I noticed that there is a make recipe to generate a coverage report, but it is somewhat broken.

It requires manually generating and updating the `cover.out` file. Thus, calling it on a new checkout fails, for example:

```
❯ make cover
go tool cover -html=cover.out
cover: open cover.out: no such file or directory
make: *** [Makefile:16: cover] Error 2
```

It is also very easy to forget about that and generate coverage reports based on an old `cover.out` that is still lying around.  Dependency tracking is the strong suit of make, so we should make use of that.

I do not know how deep you want to go into this. We could specify the whole dependency graph, but this PR contains the minimal changes to make `make cover` actually work out of the box and be up to date. The `cover.out` target is marked as phony, so it is re-generated on each call, without having to actually track all of the dependencies.

With this PR, running `make cover` always generates an up to date coverage report for the current checkout.


Makefile
```diff
@@ -12,7 +12,7 @@ fuzz:
 	done
 
 .PHONY: cover
-cover:
+cover: cover.out
```
markuswustenberg, line 15: Maybe just make cover depend on test, so that it’s always regenerated? I guess my habit is “make test cover”. :D

markuswustenberg: Ah, sorry, didn’t understand the .PHONY thing.

markuswustenberg:
Sweet, thanks!

PR #358, opened 2026-09-16
Add a gomponents skill for coding agents

maragubot:
A skill file at `skills/gomponents/SKILL.md` that teaches an agent to write gomponents well, in under 300 lines, plus a small `references/linting.md` for the dot-import whitelist. The repo doubles as a Claude Code plugin and marketplace, so it installs with `/plugin marketplace add maragudk/gomponents` and `/plugin install gomponents@gomponents`; the README says so. The plugin has no version, so every commit to main is an update, and no git tags or releases are involved.

- Covers the `Node` model, gomponents as a DSL for HTML in valid Go, the core functions with one example that uses them, common patterns, imports and package layout, `components.HTML5`, `Classes` and `JoinAttrs`, `http.Adapt`, and how to test components.
- Deliberately does not list the elements and attributes. It gives the naming rules, the element and attribute name clashes, the deprecated aliases to avoid, and tells the agent to check with `go doc`.
- Assumes the current `main`, where `JoinAttrs` flattens nested groups, so a caller's classes merge through several layers of components.

Validated over four rounds of evals, each prompt run twice, once with the skill and once without, on a fresh module. The last round scored 33 of 37 assertions with the skill against 25 without. What the skill reliably changes: components land in a package named `html` with dot imports, `Title` instead of the deprecated `TitleAttr`, `JoinAttrs` when components layer their own classes, `Iff` where a nil pointer would panic, `Adapt` with a `StatusCode()` error instead of a hand-rolled response, and tests that check behaviour instead of pinning a whole page as one string.


skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,300 @@
+---
+name: gomponents
+description: Building, rendering, or editing any HTML or UI in a Go application means using gomponents — ALWAYS invoke this skill as your FIRST action, before any Read, Write, Edit, or Bash. This fires for any view, page, layout, component, form, table, navbar, footer, or list that renders to HTML; any function returning a `Node`; anything in the `html` package; and converting HTML or templates into Go. It is a hard requirement — the user writes all HTML through gomponents, a pure-Go component library whose conventions (dot imports, `Node` composition, `Map`/`If`/`Iff`, `Group`, HTML5 documents) ordinary Go and HTML habits get wrong. Load it even when a general Go skill also applies. Skip only when no HTML or view code is touched (pure database, handler, or business logic; plain `.css`/`.js`/`.html` files; conceptual questions with no code).
```
markuswustenberg: The user may have no general Go skill

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,300 @@
...
+---
+
+# gomponents
+
+gomponents is HTML components in pure Go. A view is a Go function that returns a `Node`, and a `Node` renders itself to an `io.Writer` as HTML5. There is no template language, no code generation, and no dependencies.
```
markuswustenberg: view -> component

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,300 @@
...
+
+```sh
+go get maragu.dev/gomponents
```
markuswustenberg: maragu.dev/gomponents@latest ?

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,300 @@
...
+```
+
+The import path is `maragu.dev/gomponents`, not the GitHub URL. It needs Go 1.18 or later. The API is stable and feature-complete: new elements and attributes get added, control-flow helpers such as `IfElse` do not. Keep small helpers like that in your own project.
```
markuswustenberg: That's for contributors, not users. Leave out.

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,300 @@
...
+type Node interface {
+	Render(w io.Writer) error
+}
+```
+
+Elements, attributes, text, and groups all implement it, and all are passed as children to the same variadic element functions. Four rules explain nearly all behaviour:
+
+1. **Attributes go in the tag, everything else between the tags.** An element renders its attribute children inside the opening tag and the rest between the tags, in whatever order you pass them. By convention, write attributes first.
+2. **`nil` children are skipped.** `If` depends on this: it returns `nil` when the condition is false.
+3. **Nodes are lazy.** Building a tree captures values in closures; nothing is written until `Render` runs. Build a fresh tree per request and render it straight to the response.
```
markuswustenberg: But components are function calls, so args etc. are evaluated immediately

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,300 @@
...
+4. **Nothing is added.** No whitespace appears between nodes: `Span(Text("a")), Span(Text("b"))` renders `<span>a</span><span>b</span>`. Add `Text(" ")` where inline spacing matters.
```
markuswustenberg: Seems like a detail, do we really need that?

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,300 @@
...
+```go
+package html
+
+import (
+	. "maragu.dev/gomponents"
+	. "maragu.dev/gomponents/components"
+	. "maragu.dev/gomponents/html"
+)
+
+func Navbar(authenticated bool, currentPath string) Node {
```
markuswustenberg: I like the idea of an example component. Maybe we should make one that uses all the core features?

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,300 @@
...
+func navbarLink(href, name, currentPath string) Node {
+	return A(Href(href), Classes{"is-active": currentPath == href}, Text(name))
+}
+```
+
+## Imports and package layout
+
+Dot-import the three main packages so views read like HTML. This is the strongly preferred style unless the project already imports them another way. Whichever style the project uses, use it in every file that mentions a `Node`, handler files included: write `Node`, not `g.Node`.
+
+Put views in a package named `html`, as the gomponents example app and starter kit do, unless the project already has a views package under another name. A package named `html` can dot-import `maragu.dev/gomponents/html` without conflict.
```
markuswustenberg: don't mention example app and starter kit

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,300 @@
...
+
+Inside that package, export only what another package uses, which is usually just the pages:
```
markuswustenberg: …and fragments

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,300 @@
...
+
+- **Exported pages** are named `XxxPage` and take a props struct: `func LoginPage(props LoginPageProps) Node`.
+- **Unexported building blocks** are named in lower case after the element they wrap: `button`, `input`, `label`, `a`, `card`, `container`. Lower-case names cannot collide with the dot-imported `Button`, `Input`, `Label`, and `A`. When a block does need exporting, give it a specific name (`SubmitButton`, not `Button`) rather than dropping the dot import.
+
+`staticcheck` flags dot imports by default, so whitelist them in `.golangci.yml`:
```
markuswustenberg: This seems out of place here, maybe a separate doc for linting considerations?

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,300 @@
...
+    staticcheck:
+      dot-import-whitelist:
+        - "maragu.dev/gomponents"
+        - "maragu.dev/gomponents/components"
+        - "maragu.dev/gomponents/html"
+```
+
+The `http` package clashes with `net/http`, so alias it: `ghttp "maragu.dev/gomponents/http"`.
+
+## Core package: `maragu.dev/gomponents`
```
markuswustenberg: This table should live further up, near the example maybe?

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,300 @@
...
+| `Map(slice, func(T) Node) Group` | Turns a slice of data into nodes. Pass the result directly as a child; it is already a `Node`. |
+| `Group{...}` / `Group(nodes)` | A `[]Node` that renders as one node. Use it to return several siblings, or to pass a `children ...Node` slice on. |
+| `If(cond, node)` | `node` if `cond`, else `nil`. The node argument is evaluated either way. |
+| `Iff(cond, func() Node)` | Like `If`, but the function only runs when `cond` is true. |
+
+Element and attribute *names* are written verbatim, so they must be trusted compile-time values. Attribute *values*, `Text`, and `Textf` are escaped. Never build a name from user input, as in `Data(userKey, v)` or `El(tagFromRequest)`.
+
+Every built-in node implements `fmt.Stringer`, so `fmt.Println(node)` prints its HTML when debugging.
+
+### `If` evaluates eagerly, `Iff` doesn't
```
markuswustenberg: This is an h3 heading under h2 `Imports and package layout`, that's weird

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,300 @@
...
+There is no `IfElse`, on purpose. Use two `If`s with opposite conditions, or a function with a `switch` when the branches are more than one node:
+
+```go
+P(
+	If(ok, Text("Saved.")),
+	If(!ok, Text("Something went wrong.")),
+)
+```
+
+`Map` has no index. When you need one, keep a counter in the closure, or use the experimental `maragu.dev/gomponents/x/slices` package, whose `Map` and `Filter` pass the index and return plain slices to spread with `...`. Packages under `x/` may change without notice.
```
markuswustenberg: …although they probably won't (or something like that)

skills/gomponents/SKILL.md
maragubot: Agreed, dropping that sentence.

skills/gomponents/SKILL.md
maragubot: Agreed. Changing this one and the other places that say view: the dot-import paragraph, the package paragraph, and the Views take plain values bullet. The description keeps its list, since view is a word people type.

skills/gomponents/SKILL.md
maragubot: Yes, adding @latest.

skills/gomponents/SKILL.md
maragubot: Dropping everything after the import path sentence.

skills/gomponents/SKILL.md
maragubot: Right, lazy is the wrong word. Rewriting the rule as: building and rendering are separate. Component calls run and build the tree; nothing is written until Render runs.

skills/gomponents/SKILL.md
maragubot: Agreed, dropping the rule.

skills/gomponents/SKILL.md
maragubot: Will do. The example will cover Text, Textf, Raw, Map, Group, If, Iff, and Classes, with a short comment per feature. Attr and El stay out of it and remain in the table below.

skills/gomponents/SKILL.md
maragubot: Removing the mention.

skills/gomponents/SKILL.md
maragubot: Adding fragments, here and in the exported bullet.

skills/gomponents/SKILL.md
maragubot: Moving the whole thing, staticcheck sentence included, to references/linting.md, and leaving a one-line pointer here.

skills/gomponents/SKILL.md
maragubot: Moving the table up into the mental model section, right before the example.

skills/gomponents/SKILL.md
maragubot: Restructuring: the If/Iff and conditionals subsections fold into one Common patterns section right after the mental model, and the h3 headings go away.

skills/gomponents/SKILL.md
maragubot: Softening it to: experimental and may change, although they probably won't.

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+
+gomponents is HTML components in pure Go. A component is a Go function that returns a `Node`, and a `Node` renders itself to an `io.Writer` as HTML5. There is no template language, no code generation, and no dependencies.
+
+```sh
+go get maragu.dev/gomponents@latest
+```
+
+The import path is `maragu.dev/gomponents`, not the GitHub URL.
+
+## Mental model
```
markuswustenberg, line 15: I'm not too happy with the beginning here. Node interface is fine, but then, the three rules, which seem somewhat arbitrary. I think it's much more important to state that it's effectively a DSL for HTML that is also valid Go code, and written declaratively (although imperative components are also possible). Then after that, the core package etc.

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+1. **Attributes go in the tag, everything else between the tags.** An element renders its attribute children inside the opening tag and the rest between the tags, in whatever order you pass them. By convention, write attributes first.
+2. **`nil` children are skipped.** `If` depends on this: it returns `nil` when the condition is false.
+3. **Building and rendering are separate.** Component calls run and build the tree; nothing is written until `Render` runs. Build a fresh tree per request and render it straight to the response.
+
+The core package provides these functions. Everything else is in the `html` and `components` packages.
+
+| Function | What it does |
+|---|---|
+| `Text(s)` / `Textf(format, args...)` | HTML-escaped text. The default for all content, and the only choice for anything user-controlled. |
+| `Raw(s)` / `Rawf(format, args...)` | Unescaped text. For markup you wrote yourself: inline SVG, `<script>` and `<style>` bodies. Never for user content. |
```
markuswustenberg: "Never for unsanitized user content"

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+| `Map(slice, func(T) Node) Group` | Turns a slice of data into nodes. Pass the result directly as a child; it is already a `Node`. |
```
markuswustenberg: What does the "Pass the result…" mean?

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+| `Group{...}` / `Group(nodes)` | A `[]Node` that renders as one node. Use it to return several siblings, or to pass a `children ...Node` slice on. |
+| `If(cond, node)` | `node` if `cond`, else `nil`. The node argument is evaluated either way. |
+| `Iff(cond, func() Node)` | Like `If`, but the function only runs when `cond` is true. |
+| `El(name, children...)` | Any element. Use it for elements the `html` package lacks: custom elements, or SVG children such as `El("path", Attr("d", "..."))`. |
+| `Attr(name)` / `Attr(name, value)` | Boolean or valued attribute. Use it for attributes the `html` package lacks: `Attr("hx-get", "/items")`, `Attr("onclick", "...")`. More than one value panics. |
+
+Element and attribute *names* are written verbatim, so they must be trusted compile-time values. Attribute *values*, `Text`, and `Textf` are escaped. Never build a name from user input, as in `Data(userKey, v)` or `El(tagFromRequest)`.
+
+Every built-in node implements `fmt.Stringer`, so `fmt.Println(node)` prints its HTML when debugging.
```
markuswustenberg: Move further down or drop

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+
+One component using all of the above except `El` and `Attr`:
```
markuswustenberg: add the word "example"

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+	)
+}
+```
+
+## Common patterns
+
+- **`If` builds its node before checking the condition.** `If(user != nil, Text(user.Name))` panics when `user` is nil, because `user.Name` runs first. Use `Iff` when the node reads through a pointer, indexes a slice, or does work worth skipping.
+- **There is no `IfElse`.** Use two `If`s with opposite conditions, or a function with a `switch` when the branches are more than one node.
+- **`Map` has no index.** When you need one, keep a counter in the closure, or use `maragu.dev/gomponents/x/slices`, whose `Map` and `Filter` pass the index and return plain slices to spread with `...`. Packages under `x/` are experimental and may change, although they probably won't.
+- **Building blocks take `children ...Node`** and pass them on with `Group(children)`. Groups are transparent, so a caller's attributes (`ID`, `Attr("hx-target", ...)`) are placed on the root element: `input(Type("email"), Name("email"), Required())` needs no parameters of its own.
```
markuswustenberg: Keep HTMX out of this

markuswustenberg: Also, I don't understand the `input` example

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+- **Placement classes come from the caller.** A block used in several places takes a `class string` first and joins it with its own classes, as in the example below. Use `JoinAttrs` instead when callers pass attribute nodes rather than a string.
```
markuswustenberg: Actually I prefer the JoinAttrs approach, not the `class string` one

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+- **One look, several elements: pass the element function.** `heading(H1, "", Text(title))` and `heading(Dt, "", Text(name))` share one style.
```
markuswustenberg: Don't understand this one

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+- **Many options: use a props struct**, as `HTML5Props` does, rather than a long parameter list.
```
markuswustenberg: Drop, up to user style

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+- **Components take plain values.** Load data and handle errors before calling the component; `Render` fails only on a write error, so a component has no error path.
```
markuswustenberg: drop?

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+- **Dynamic attributes** work like dynamic elements, because `nil` is skipped inside the tag too: `If(disabled, Disabled())`, `Classes{...}`, `Value(u.Email)`.
+- **Fragments** for htmx and similar are blocks rendered without the layout. Return a `Group` when a fragment has several root elements.
```
markuswustenberg: drop htmx reference, just HTML fragments

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+- **`<script>` and `<style>` bodies need `Raw`.** `Text` would turn `&&` into `&amp;&amp;` and break the code. Keep user data out of those bodies; pass it through `Data` attributes instead, which are escaped and unescaped by the browser.
+- **Two `Class` calls make two attributes**, and the browser keeps only the first. Combine them into one string, use `Classes`, or use `JoinAttrs`.
+- **`Classes` sorts.** `Classes{"tier": true, "popular": true}` renders `class="popular tier"`. Fine for CSS, but a test that pins the exact string must expect the sorted order.
+- **Void elements silently ignore non-attribute children.** `Img(Text("x"))` renders `<img>` with no error.
+- **`Attr` with two or more values panics** at construction time, not at render time.
```
markuswustenberg: That's a general rule (do we have it anywhere else?): bugs make a panic, should be caught in testing

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+
+```go
```
markuswustenberg: This example comes kind of out of the blue. Why is it here?

skills/gomponents/SKILL.md
maragubot: Agreed. Replacing the rules with a paragraph on gomponents as a DSL for HTML in valid Go, written declaratively, with a short HTML-to-Go contrast, then the core table and the example. Only the attribute-placement point survives, as a sentence.

skills/gomponents/SKILL.md
maragubot: Changing to that wording.

skills/gomponents/SKILL.md
maragubot: Dropping the sentence; the signature and the example already show it.

skills/gomponents/SKILL.md
maragubot: Moving it to Common patterns as a bullet.

skills/gomponents/SKILL.md
maragubot: Adding it.

skills/gomponents/SKILL.md
maragubot: Dropping the htmx mention, and rewriting the input clause to say what it meant: a block that only adds classes to an element needs no parameters, since callers pass attributes as children and they land on the element.

skills/gomponents/SKILL.md
maragubot: Making JoinAttrs the one way and dropping the class string helper and its snippet.

skills/gomponents/SKILL.md
maragubot: Dropping it.

skills/gomponents/SKILL.md
maragubot: Dropping it.

skills/gomponents/SKILL.md
maragubot: Dropping it.

skills/gomponents/SKILL.md
maragubot: Removing the htmx reference.

skills/gomponents/SKILL.md
maragubot: Dropping the bullet and stating the general rule once in the testing section: programming errors panic when the component is built, so rendering each component in a test catches them.

skills/gomponents/SKILL.md
maragubot: Removed along with the two bullets it illustrated.

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+
+# gomponents
+
+gomponents is HTML components in pure Go. A component is a Go function that returns a `Node`, and a `Node` renders itself to an `io.Writer` as HTML5. There is no template language, no code generation, and no dependencies.
+
+```sh
+go get maragu.dev/gomponents@latest
+```
+
+The import path is `maragu.dev/gomponents`, not the GitHub URL.
```
markuswustenberg: Maybe just drop this? Unlikely, enforced by compiler since in go.mod

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+
+Everything is a `Node`:
+
+```go
+type Node interface {
+	Render(w io.Writer) error
+}
+```
+
+Elements, attributes, and text all implement it, and all are passed as children to the same variadic element functions. The result is a DSL for HTML that is also valid Go: an element function takes children, an attribute function takes a value, and nesting calls mirrors nesting tags. Attribute children go in the tag and everything else goes between the tags, whatever order you pass them in; by convention, write attributes first.
```
markuswustenberg: "like a DSL"

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+
+```html
+<a href="/about" class="nav-link">About</a>
+```
+
+```go
+A(Href("/about"), Class("nav-link"), Text("About"))
+```
+
+You write it declaratively, but a component is an ordinary function and can be as imperative as it needs to be.
```
markuswustenberg: "The idiomatic way is to write it declaratively, but…" ?

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+
+The core package provides these functions. Everything else is in the `html` and `components` packages.
+
+| Function | What it does |
+|---|---|
+| `Text(s)` / `Textf(format, args...)` | HTML-escaped text. The default for all content, and the only choice for anything user-controlled. |
```
markuswustenberg: add "unsanitized" here as well? Clearly not the only choice.

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+type User struct {
+	Name   string
+	Unread int
+}
+
+// Navbar shows the site links and, for a logged-in user, their name and a log out link.
+// user is nil for a visitor.
+func Navbar(links []NavLink, currentPath string, user *User) Node {
+	return Nav(Class("navbar"),
+		// Map turns a slice into nodes, one per element.
```
markuswustenberg: Maybe drop the inline comments here, since already given in the table above?

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+- **`Map` has no index.** When you need one, keep a counter in the closure, or use `maragu.dev/gomponents/x/slices`, whose `Map` and `Filter` pass the index and return plain slices to spread with `...`. Packages under `x/` are experimental and may change, although they probably won't.
+- **Building blocks take `children ...Node`** and pass them on with `Group(children)`. Groups are transparent, so a caller's attributes (`ID`, `Name`) are placed on the root element. A block like `input(children ...Node)` that only adds classes to `Input` needs no parameters, because callers pass `Type("email")`, `Name("email")`, or `Required()` as children and they land on the input.
+- **Callers add classes through `JoinAttrs`.** A block joins its own `Class` with whatever the caller passes, so `card(Class("mt-4"), ...)` renders one `class` attribute. See the `components` package below.
+- **Dynamic attributes** work like dynamic elements, because `nil` is skipped inside the tag too: `If(disabled, Disabled())`, `Classes{...}`, `Value(u.Email)`.
+- **Fragments** are components rendered without the layout. Return a `Group` when a fragment has several root elements.
```
markuswustenberg: Can we somehow say that "fragments" is just a commonly used term, like "component" or "view", and nothing special in gomponents? But concisely.

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+- **Print a node to see its HTML.** Every built-in node implements `fmt.Stringer`, so `fmt.Println(node)` works when debugging.
+- **`<script>` and `<style>` bodies need `Raw`.** `Text` would turn `&&` into `&amp;&amp;` and break the code. Keep user data out of those bodies; pass it through `Data` attributes instead, which are escaped and unescaped by the browser.
```
markuswustenberg: +unsanitized

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+- **Two `Class` calls make two attributes**, and the browser keeps only the first. Combine them into one string, use `Classes`, or use `JoinAttrs`.
```
markuswustenberg, line 106: We should make a note further up that gomponents does not enforce valid HTML, as a design decision, just like when writing raw HTML yourself.

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+- **`Classes` sorts.** `Classes{"tier": true, "popular": true}` renders `class="popular tier"`. Fine for CSS, but a test that pins the exact string must expect the sorted order.
```
markuswustenberg: Maybe we should drop this? Too detailed?

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+## Imports and package layout
+
+Dot-import the three main packages so components read like HTML. This is the strongly preferred style unless the project already imports them another way. Whichever style the project uses, use it in every file that mentions a `Node`, handler files included: write `Node`, not `g.Node`.
+
+Put components in a package named `html`, unless the project already has one under another name. A package named `html` can dot-import `maragu.dev/gomponents/html` without conflict.
+
+Inside that package, export only what another package uses, which is usually just the pages and fragments:
+
+- **Exported pages and fragments** are named for what they are and take a props struct: `func LoginPage(props LoginPageProps) Node`, `func UserRow(u User) Node`.
+- **Unexported building blocks** are named in lower case after the element they wrap: `button`, `input`, `label`, `a`, `card`, `container`. Lower-case names cannot collide with the dot-imported `Button`, `Input`, `Label`, and `A`. When a block does need exporting, give it a specific name (`SubmitButton`, not `Button`) rather than dropping the dot import.
```
markuswustenberg: Well, actually not lowercase but camelCase with lower first letter

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+
+The `http` package clashes with `net/http`, so alias it: `ghttp "maragu.dev/gomponents/http"`.
```
markuswustenberg: "often clashes"

markuswustenberg: So consider aliasing it

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+
+For linter setup, see `references/linting.md`.
+
+## The `html` package: elements and attributes
+
+Element functions take `...Node`. Attribute functions take one `string` value, or nothing for boolean attributes (`Required()`, `Disabled()`, `Checked()`). All values are strings, so convert numbers yourself: `Width(strconv.Itoa(w))`.
```
markuswustenberg: There's also one attribute that takes variadic string, since it can be both bool and key-value attr

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+|---|---|---|
+| `cite` | `Cite` | `CiteAttr` |
+| `data` | `DataEl` | `Data(name, value)`, renders `data-<name>` |
+| `form` | `Form` | `FormAttr` |
+| `label` | `Label` | `LabelAttr` |
+| `slot` | `SlotEl` | `SlotAttr` |
+| `style` | `StyleEl` | `Style` |
+| `title` | `TitleEl` | `Title` |
+
+`CiteEl`, `DataAttr`, `FormEl`, `LabelEl`, `StyleAttr`, and `TitleAttr` still compile but are deprecated; do not use them in new code.
```
markuswustenberg: drop after `;`

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+
+`Data("id", v)` renders `data-id="..."` and `Aria("label", v)` renders `aria-label="..."`. `Doctype(node)` prefixes `<!doctype html>`; `HTML5` already calls it.
```
markuswustenberg: HTML5 is not in the `html` package

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+
+When unsure whether a helper exists or how it is cased, check rather than guess:
+
+```sh
+go doc maragu.dev/gomponents/html | grep -i colspan
+```
+
+If nothing turns up, `El` and `Attr` produce the same output a dedicated helper would. Left out on purpose: SVG child elements, `on*` event handler attributes, and framework attributes. For htmx and Datastar there are typed helpers in `maragu.dev/gomponents-htmx` and `maragu.dev/gomponents-datastar`; otherwise `Attr("hx-get", ...)` is fine.
```
markuswustenberg: Drop HTMX and Datastar refs. Maybe drop the whole "left out on purpose" stuff

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+
+## The `components` package
+
+**`HTML5`** renders a complete document: doctype, `<html>`, a `<head>` with charset, viewport, title, and optional description, and the `<body>`. Attribute nodes in `Body` or `Head` are placed on that element. Every app has one layout function like this, and every page calls it:
```
markuswustenberg: Most apps?

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+```go
+func page(title string, body ...Node) Node {
+	return HTML5(HTML5Props{
+		Title:    title + " - MyApp",
+		Language: "en",
+		Head: Group{
+			Link(Rel("stylesheet"), Href("/static/app.css")),
+			Script(Src("/static/app.js"), Defer()),
+		},
+		Body: Group{Class("min-h-full bg-gray-50"),
```
markuswustenberg, line 165: Also show HTMLAttrs?

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+// <button class="btn-primary mt-4 btn">Save</button>
+```
+
+## The `http` package
+
+`Adapt` turns a handler that returns `(Node, error)` into an `http.HandlerFunc`:
+
+- The node is rendered even when there is an error, so return an error page along with the error.
+- If the error has a `StatusCode() int` method, that status is sent; any other error sends 500.
+- A `nil` node writes nothing; return it after you have issued a redirect yourself.
```
markuswustenberg: Or when there's an error and its message should be shown instead

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+Put components in a package named `html`, unless the project already has one under another name. A package named `html` can dot-import `maragu.dev/gomponents/html` without conflict.
+
+Inside that package, export only what another package uses, which is usually just the pages and fragments:
+
+- **Exported pages and fragments** are named for what they are and take a props struct: `func LoginPage(props LoginPageProps) Node`, `func UserRow(u User) Node`.
+- **Unexported building blocks** are named in lower case after the element they wrap: `button`, `input`, `label`, `a`, `card`, `container`. Lower-case names cannot collide with the dot-imported `Button`, `Input`, `Label`, and `A`. When a block does need exporting, give it a specific name (`SubmitButton`, not `Button`) rather than dropping the dot import.
+
+The `http` package clashes with `net/http`, so alias it: `ghttp "maragu.dev/gomponents/http"`.
+
+For linter setup, see `references/linting.md`.
```
markuswustenberg: Add this in the testing section instead

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+	if err != nil {
+		return html.ErrorPage(), err
+	}
+	return html.UserPage(u), nil
+}))
+```
+
+## Testing components
+
+Test exported pages and components: one `TestComponent` function per exported component, with subtests for one happy path, the error cases, and the edge cases. Test what matters in each: the branch that depends on input, the value that must be escaped, the link that appears for one kind of user and not another. An expected string for a whole page restates the component and breaks on every unrelated change, so reserve exact-string comparison for small blocks and fragments, even when converting existing HTML. Programming errors in a component, such as `Attr` given two values, panic when the component is built, so rendering each component in a test catches them. gomponents has no public test helpers, so render to a `strings.Builder` through a small helper and check with `strings.Contains`:
```
markuswustenberg: Test what matters in each: for example, …

markuswustenberg: Maybe we should file an issue to add test helpers?

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+	if err := n.Render(&b); err != nil {
+		t.Fatal(err)
+	}
+	return b.String()
+}
+```
+
+Test handlers through `Adapt` with `httptest.NewRecorder`, and assert on the status code and a few body markers.
+
+Unexported blocks are tested from inside the package. To reach them from an external `_test` package, re-export them in an `export_internal_test.go` file with `var Card = card`. That re-export collides with dot-imported element names just as an exported function would, so it works for `card` but not for `button`.
```
markuswustenberg: …or use internal tests, according to project preference.

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Further reading
+
+- API reference: https://pkg.go.dev/maragu.dev/gomponents
```
markuswustenberg: Use go doc instead?

skills/gomponents/SKILL.md
```diff
@@ -0,0 +1,262 @@
...
+- Guides: https://www.gomponents.com
```
markuswustenberg: Drop, this is the guide

README.md
```diff
@@ -73,6 +73,15 @@ func NavbarLink(href, name, currentPath string) Node {
 For a more complete example, see [the examples directory](internal/examples/).
 There's also the [gomponents-starter-kit](https://github.com/maragudk/gomponents-starter-kit) for a full application template.
 
+### Coding agents
+
+There's a [skill](skills/gomponents/SKILL.md) that teaches coding agents how to use gomponents. Install it as a plugin in Claude Code:
```
markuswustenberg: …or Codex (find Codex instructions, hide both behind `<details>`)

README.md
```diff
@@ -73,6 +73,15 @@ func NavbarLink(href, name, currentPath string) Node {
 For a more complete example, see [the examples directory](internal/examples/).
 There's also the [gomponents-starter-kit](https://github.com/maragudk/gomponents-starter-kit) for a full application template.
 
+### Coding agents
+
```
markuswustenberg, line 77: Also note that this is a good, concise introduction to gomponents for humans.

.claude-plugin/marketplace.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "gomponents",
+  "owner": {
```
markuswustenberg, line 3: Can we skip the owner entirely? Is it required?

skills/gomponents/SKILL.md
maragubot: Dropping it.

skills/gomponents/SKILL.md
maragubot: Changing to: like a DSL for HTML, in valid Go.

skills/gomponents/SKILL.md
maragubot: Using that phrasing.

skills/gomponents/SKILL.md
maragubot: Changing to: the recommended choice for unsanitized user content.

skills/gomponents/SKILL.md
maragubot: Dropping the comments that restate the table, keeping the one about user.Name and nil, which is specific to this code.

skills/gomponents/SKILL.md
maragubot: Rewording to: a fragment is just a component returned without the layout; gomponents has no special notion of it.

skills/gomponents/SKILL.md
maragubot: Adding unsanitized.

skills/gomponents/SKILL.md
maragubot: Adding a sentence to the mental model: gomponents does not check that the result is valid HTML, by design, just as writing HTML by hand doesn't.

skills/gomponents/SKILL.md
maragubot: Dropping it.

skills/gomponents/SKILL.md
maragubot: Correcting both mentions to camelCase with a lower-case first letter.

skills/gomponents/SKILL.md
maragubot: Changing to: often clashes, so consider aliasing it.

skills/gomponents/SKILL.md
maragubot: Moving the pointer to the end of the testing section.

skills/gomponents/SKILL.md
maragubot: Adding: a few take ...string, for attributes that are valid both with and without a value.

skills/gomponents/SKILL.md
maragubot: Dropping it.

skills/gomponents/SKILL.md
maragubot: Dropping the Doctype sentence; the components section already covers the doctype.

skills/gomponents/SKILL.md
maragubot: Keeping only the El/Attr fallback sentence; the left-out list and the htmx/Datastar references go.

skills/gomponents/SKILL.md
maragubot: Changing to: most apps have one layout function like this.

skills/gomponents/SKILL.md
maragubot: Adding HTMLAttrs, and Description too, so the example shows all six HTML5Props fields.

skills/gomponents/SKILL.md
maragubot: Rewriting the bullets: pages return an error page with the error (403, 404, 500), fragments commonly return nil, err, which sends only the status, and a nil node also follows a response you wrote yourself, such as a redirect.

skills/gomponents/SKILL.md
maragubot: Rewording to 'for example …'. Filed #359 for public test helpers.

skills/gomponents/SKILL.md
maragubot: Making internal tests and the re-export equal options, by project preference.

skills/gomponents/SKILL.md
maragubot: Switching to go doc.

skills/gomponents/SKILL.md
maragubot: Dropping it.

README.md
maragubot: Adding the Codex commands and collapsing both behind details blocks.

README.md
maragubot: Adding that sentence.

.claude-plugin/marketplace.json
maragubot: Checked with claude plugin validate: owner is required (validation fails without it), but the email is not. Keeping the name only, and dropping the email from the plugin author too.

PR #364, opened 2026-09-17
Return shared nodes from childless element and boolean attribute helpers

maragubot:
Every element helper in `html/elements.go` except `html.Doctype` (114 of them) now returns a shared, package-level node when called with no children, and every zero-argument boolean attribute helper in `html/attributes.go` (17 of them) returns one always. Those calls go from 1 allocation to 0.

- Each singleton sits directly below its helper as `var <tag>El = g.El("<tag>")` / `var <name>Attr = g.Attr("<name>")`. The deprecated shims `html.CiteEl`, `html.FormEl` and `html.LabelEl` already delegate to `html.Cite`, `html.Form` and `html.Label`, so they reach the same node without a second guard.
- Rendered output is byte-for-byte identical. Nodes with children or values take the old path unchanged, including `html.Div(nil)`, `html.Div(g.If(false, x))` and `html.Div(g.Group{})`, which all have one child.
- Sharing is safe: a node is a closure over an immutable string and a nil slice, and `g.Node` exposes only `Render`. Node identity was never observable, since comparing two `g.Node` values holding func types panics.
- The core package, `components/`, `http/` and `html.Doctype` are untouched. No new exported API.
- New `BenchmarkElement` and `BenchmarkBooleanAttribute` in `html/benchmark_test.go`, and two matching rows in the README performance table.

## Benchmarks

`go test -bench . -benchmem -count=10 ./...` on an Apple M4, before from a pristine export of `main` with the two new benchmarks copied in, compared with `benchstat`.

```
                                                 │  before.txt  │              after.txt              │
                                                 │    sec/op    │   sec/op     vs base                │
Element/element_without_children-10                12.965n ± 1%   5.884n ± 1%  -54.62% (p=0.000 n=10)
BooleanAttribute/boolean_attribute-10               9.669n ± 0%   4.768n ± 0%  -50.69% (p=0.000 n=10)
RealisticPage/construct_and_render/discarded-10     67.41µ ± 0%   67.14µ ± 1%   -0.39% (p=0.023 n=10)
RealisticPage/construct_and_render/buffered-10      90.72µ ± 1%   89.81µ ± 0%   -1.00% (p=0.000 n=10)
RealisticPage/render_pre-built_tree/discarded-10    43.28µ ± 1%   43.14µ ± 0%   -0.33% (p=0.007 n=10)
RealisticPage/render_pre-built_tree/buffered-10     66.72µ ± 0%   66.33µ ± 0%   -0.59% (p=0.004 n=10)

                                                 │  before.txt  │                 after.txt                 │
                                                 │     B/op     │     B/op      vs base                     │
Element/element_without_children-10                  48.00 ± 0%      0.00 ± 0%  -100.00% (p=0.000 n=10)
BooleanAttribute/boolean_attribute-10                24.00 ± 0%      0.00 ± 0%  -100.00% (p=0.000 n=10)
RealisticPage/construct_and_render/discarded-10    126.6Ki ± 0%   126.6Ki ± 0%    -0.04% (p=0.000 n=10)
RealisticPage/construct_and_render/buffered-10     126.6Ki ± 0%   126.6Ki ± 0%    -0.04% (p=0.000 n=10)
RealisticPage/render_pre-built_tree/discarded-10   1.375Ki ± 0%   1.375Ki ± 0%         ~ (p=1.000 n=10)
RealisticPage/render_pre-built_tree/buffered-10    1.375Ki ± 0%   1.375Ki ± 0%         ~ (p=1.000 n=10)

                                                 │ before.txt  │                after.txt                 │
                                                 │  allocs/op  │  allocs/op   vs base                     │
Element/element_without_children-10                 1.000 ± 0%    0.000 ± 0%  -100.00% (p=0.000 n=10)
BooleanAttribute/boolean_attribute-10               1.000 ± 0%    0.000 ± 0%  -100.00% (p=0.000 n=10)
RealisticPage/construct_and_render/discarded-10    3.044k ± 0%   3.042k ± 0%    -0.07% (p=0.000 n=10)
RealisticPage/construct_and_render/buffered-10     3.044k ± 0%   3.042k ± 0%    -0.07% (p=0.000 n=10)
RealisticPage/render_pre-built_tree/discarded-10    18.00 ± 0%    18.00 ± 0%         ~ (p=1.000 n=10)
RealisticPage/render_pre-built_tree/buffered-10     18.00 ± 0%    18.00 ± 0%         ~ (p=1.000 n=10)
```

The realistic page contains exactly two childless nodes, so it drops from 3044 to 3042 allocations. Real, reproducible, and worth nothing; the per-call numbers are the story.

### Controls

`BenchmarkEl` and `BenchmarkAttr` in the core package are the controls, and the two full runs disagreed by a couple of percent on wall time even though `gomponents.go` and `gomponents_benchmark_test.go` are byte-identical in both trees:

```
El/normal_elements-10                               17.27n ± 1%   16.93n ± 1%   -1.94% (p=0.005 n=10)
Attr/construct_and_render/discarded/boolean-10      11.06n ± 0%   10.66n ± 1%   -3.57% (p=0.000 n=10)
```

Re-measuring just those two trees back to back settles it as between-run drift on this machine, not a real change:

```
                      │ ctrl-baseline.txt │         ctrl-worktree.txt          │
El/normal_elements-10         15.84n ± 1%   15.99n ± 2%  +0.95% (p=0.014 n=10)
Raw/raw_element-10            2.784n ± 0%   2.767n ± 1%  -0.63% (p=0.002 n=10)
geomean                       6.642n        6.653n       +0.16%
```

Bytes and allocations per op for the controls are identical in every comparison.

## Notes for review

- The `Realistic full page` row in the README is left at 3044 allocations. The branch produces 3042, which is not a meaningful move, and that row's nanoseconds already came from an older run than either of these. Say the word and I will refresh the whole row.
- `sink` in `html/elements_test.go` is a package-level variable that nothing reads. It keeps `testing.AllocsPerRun` honest: without it the compiler could drop the call being measured and the zero-allocation assertion would pass for the wrong reason.
- `docs/diary/2026-09-17-html-singletons.md` has the full narrative, including a correction to Step 1: it lists `Data`/`DataEl` as a pair of helpers rendering the same tag, but `html.Data` is the data-attribute function, not an element.


markuswustenberg:
Note that this is an experiment, not sure yet whether it's worth doing.

